PR #26, opened 2019-06-21
Adds new component that allows user to specify .spiceinit contents.

### LINKED ISSUE #25: Support specifying user-defined .spiceinit in the schematic.
Hi! I've been using your software, it's been really useful when simulating SPICE models. 

I've run into issues when using PSPICE models. ngspice, in fact, supports PSPICE models, but it requires a setting in .spiceinit (namely "set ngbehavior=ps"). The ngspice simulator class seems to create an empty (when not using xspice) spiceinit each time it runs a simulation, so it does effectively prevent users from overriding ngbehavior.

One possible way would be to have a new component ("SPICEINIT") with a text property containing text to be included in the spiceinit file created by the simulator class. I think this solution fits nicely the use case, because enabling PSPICE behavior should be done on a per-schematic basis, as it depends on which models are included into it.

I'd like to send you a PR if you agree with the general approach. WDYT?

spueblas:
Adds a new SPICE-only component (SPICEINIT), that users can use to specify .spiceinit contents. For this component, a new dialog was created, which allows editing a single text box (used in this case for the contents of the spiceinit file), and could potentially be reused for other purposes. 

The new component, SPICEINIT, is only used in the ngspice simulator class. Other simulators, like Xyce, ignore this new component. For ngspice simulator, the specified contents of spiceinit are forwarded to the XSPICE_CMbuilder class, which is currently in charge of creating the spiceinit file. This is a bit unfortunate, as it would be cleaner if the responsibility of creating spiceinit was kept by Ngspice class.

Still to do, create an icon for the new component as it currently is reusing sp_options.png. BTW, how do you create those icons? I'm afraid if I try to do it manually with GIMP or something it will not look the same style as the other icons.

Screenshot displaying how it shows in the component selector and the schematics:
Screenshot from 2019-06-21 17-09-53
Fixes #25


ra3xdh:
Looks good, surely it could be useful. I will try to test it in the next days.

ra3xdh:
> BTW, how do you create those icons?

I open an exisitng icon with KolorPaint or another such software and edit it. Yes can use .OPTIONS icon and replace the text. GIMP is overkill for such task.

spueblas:
Added an icon for the new component. PTAL.

ra3xdh:
I have tested it. The XSPICE compatibility seems not to be broken. I have found no further problems with the new component.

ra3xdh:
Are you planning to add some new features for this `.spinit` device? Otherwise I will merge this PR.

spueblas:
I was not planning to add any new features for this new component.

PR #27, opened 2019-06-23
Correct input pins of the opamp component by swapping them.

### LINKED ISSUE #20: ngspice op amp model clipping not working correctly
Hello! It looks like the op amp clipping isn't working correctly during saturation, at least using the ngspice backend. It seems to do this weird thing where the output gets stuck to the voltage rails instead of returning to normal operation. Also the op amp seems to saturate in the wrong direction somehow; when the clipping starts it jumps to the positive rail instead of saturating on the negative one. Oddly enough, the clipping seems to work correctly if I flip the positive and negative inputs, and the op amp works when it's not clipping with the inputs plugged in either way. Maybe the netlist generator has the two inputs mixed up or something? I've attached a simple test circuit below:

[test.zip](https://github.com/ra3xdh/qucs_s/files/2688208/test.zip)

spueblas:
Fixes #20.

### DIFF
--- a/qucs/components/opamp.cpp
+++ b/qucs/components/opamp.cpp
@@ -76,10 +76,10 @@ Element* OpAmp::info(QString& Name, char* &BitmapFile, bool getNewOne)
 
 QString OpAmp::spice_netlist(bool isXyce)
 {
-    QString in_p = Ports.at(0)->Connection->Name;
-    if (in_p=="gnd") in_p="0";
-    QString in_m = Ports.at(1)->Connection->Name;
+    QString in_m = Ports.at(0)->Connection->Name;
     if (in_m=="gnd") in_m="0";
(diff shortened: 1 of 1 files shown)


ra3xdh:
Thank you finding this! But the bug should be corrected in auto-generated SPICE code, because the simulation using Qucsator backend is correct. Look at the `Opamp::spice_netlist` method. Probably the `in_p` and `in_m` should be swapped here.

spueblas:
Oops, you're correct. I fixed it in the PR. I tested it works correctly with ngspice and qucsator.

Thanks!

ra3xdh:
Now everything should be correct. Please squash two commits in one using interactive rebase `rebase -i` and force push in your branch and then I will merge the PR.

spueblas:
Done.

PR #29, opened 2019-11-18
Added four terminal NMOS and PMOS symbols to the SpiceLibrary file defaults

yrrapt:
Allows for spice MOS subcircuits (ie. spice models from PDKs where parasitic diodes are modelled) to be easily used in QUCS-S.

Symbols copied from the non-linear components library


ra3xdh:
Thank you! I have merged this.

PR #64, opened 2021-06-10
Update qucs2spice.cpp

bjmuld:
Ensures Qucs components print to seperate lines even if required parameters are missing...


bjmuld:
Sorry,   did this from the Github gui and see that it's based on `master` and not `current`.  Will re-submit

PR #91, opened 2022-05-29
use QJsonDocument instead of QScriptEngine

Zergud:


ra3xdh:
Thanks for providing a patch! I will test it with my projects using Verilog-A modules and share more feedback in the next few days.

ra3xdh:
I have tested this patch with my old projects using Verilog-A devices and found no problem. The symbols could be created and loaded. I have merged the patch into the `current` branch and removed remaining QtScript references from CMake and some header files by commit https://github.com/ra3xdh/qucs_s/commit/dc04a935d8e849692131acd05efc70ace2cd2f1a

PR #102, opened 2022-07-05
Fix source typo

luzpaz:
This may also be a bugfix. Please review.


ra3xdh:
For C++ the argument name in declaration and implementation may be different and even empty. So, it is not a bug. But the different names may be misleading. Thanks for finding this!

PR #112, opened 2022-07-09
Added OpenSUSE Tumbleweed package names to README.md

powerjungle:
Also reorganized the file for clarity.


ra3xdh:
Thanks for updating the build instructions! It would be nice also to specify dependencies for Fedore/CentOS (other RPM-based).

powerjungle:
Yeah, but this was the only one I could test directly.

Issue #142, opened 2022-08-25
Missing tools in Tools menu


ra3xdh:
I have noticed the following problems with the tools:

- [x] Qucs-attenuator is missing in binary distribution but present in the source tree
- [x] Qucs-transcalc is missing in source tree; it make sense to backport it from Qt4 Qucs
- [x] Qucs-Rescodes was removed from Qucs and Qucs-S distributions around  2016 but the menu point still present
- [x] Qucs-powercombining is missing; backport from Qt4 Qucs

ra3xdh:
I have imported transcalc and removed remains of rescodes by the recent commit on `current` branch. Closing this.

ra3xdh:
The powercombining tool is also missing. It's need to port it to Qt5 and include to the Qucs-S distribution.

ra3xdh:
I have backported the powercombining tool from the latest Qucs Qt4 git. Now it should be available in the next Qucs-S release.
![image](https://user-images.githubusercontent.com/4920080/192145948-f39b2321-ce6e-4539-a19d-d2b5fc786dfb.png)

ra3xdh:
All missing tools are restored now. Closing this as completed.

Issue #78, opened 2022-03-02
Provide proper tutorial and show warning when the schematic contain only DC simulation


tckosvic:
Compiled qucs-s 0.23 and ran simple tutorial #1 (voltage divider) after creating schematic.  Looked to run ok.  Simulation dialogue showed:

`Ngspice started...

No compatibility mode selected!

Circuit: * qucs 0.0.23 /home/tom/.qucs/tuts_prj/dc-1.sch

ngspice-36 done
Simulation finished`

Data page *.dpl is created.  Pasting a cartesian table into that page as per tutorial shows no results just blank fields.

The AppImage file for qucs-s 0.23 also runs problem but presents no results.

I have plain qucs 0.19 linux compiled on my machine.  That runs divider tutorial fine and shows results.
I also have windows qucs 0.19 in wine that also runs and presents results.

Any ideas as to cause of problem?

tom kosvic

EDIT from @ra3xdh : 
The list of actions to prevent such issues in the future:

- [x] Replace Qucs tutorial by Qucs-S tutorial
- [x] Replace tutorial link in the Help menu by Qucs-S tutorial, not Qucs tutorial.
- [x] Show a warning when only DC simulation is presented in the circuit; this simulation has no effect with Ngspice/Xyce;

ra3xdh:
Hello Tom,
it is not a bug in Qucs-S. The DC simulation has no effect with Ngspice. You need to add TRAN simulation to see the result. Look a the example at the `ngspice` subdirectory. The Qucs tutorial partially is not valid when the Ngspice is used as the default simulation kernel and need to be updated.

ra3xdh:
Probably it make sense to give a warning when only DC simulation found in the schematic. For Ngspice/Xyce it has no effect.

ra3xdh:
The list of actions to prevent such issues in the future:

- [ ] Update Chapter 1 of the documentation and provide proper tutorial; don't use tutorial from Qucs;
- [ ] Show a warning when only DC simulation is presented in the circuit; this simulation has no effect with Ngspice/Xyce;

ra3xdh:
The links in the Help menu are pointing now only to to the Qucs-S tutorials and help documents. The links to old Qucs help are removed because otherwise it can be misleading for new users.

Issue #54, opened 2020-09-29
qucsator and unforking


felix-salfelder:
(Not an issue, but an announcement)

I have moved Qucsator into a separate repository within the Qucs project. It should be shared between qucs-s and qucs.

You will need the equivalent of https://github.com/Qucs/qucs/pull/1015, and perhaps check if there are any deviations that need to be considered. The rest of Qucs-s should be ported back to Qucs once Qt5 is working sufficiently.

ra3xdh:
Hello Felix,

The Qucs-S doesn't contain qucsator by itself. So, the Qucsator could be installed in the system independently and shared by several frontends. I have recently evaluated the `modular` branch of the Qucs project and I don't consider the unforking in the near future. The `modular` branch still be in very early development stage and has many incompatibilities with Qucs-S. And it's not clear for me how to resolve it. Currently the software architecture of Qucs-S is monolithic and full of legacy code and may be far from ideal, but it is stable and ready for the practical usage. I cannot say this about the `modular` branch. 

Best regards,
Vadim

felix-salfelder:
On Sun, Mar 20, 2022 at 05:27:35AM -0700, Vadim Kusnetsov wrote:
> The Qucs-S doesn't contain qucsator by itself. So, the Qucsator could
> be installed in the system independently and shared by several
> frontends.

Thanks. The latest Qucsator release has critical bugs that make it
unusable on some platforms. A maintainer is required to eventully leave
a functioning program to posterity.

> The `modular` branch [..] has many incompatibilities with Qucs-S. And
> it's not clear for me how to resolve it.

The starting point is a spice netlister and a (ng?) spice driver. This
will be very similar to the qucsator netlister and qucsator driver (with
some executable name and syntax replaced respectively). Possible next
step will be a "raw" file parser plugin. Currently, there are plugins
for reading and writing .dat files only -- but they give the idea.

Whoever does it will receive help and probably learn quite a lot.

PR #149, opened 2022-09-11
Restore missing default Qucs tools

polyelectronics:
Restore missing Qucs default tools as well changes are made to avoid warnings
from the g++ compiler.


ra3xdh:
Sorry, I cannot accept this PR. The reasons are the following;

* You are reverting `qucs-core` back to the source tree. The qucsator is a separate package maintained by another team https://github.com/qucs/qucsator It should not be merged with GUI package. See also #54
* The same for qucsconv because it's a part of qucsator package 
* The qucsator test suite must be distributed with qucsator package but not with GUI package. 
* The backport of qucs-powercoming makes sense, but must be in a separate PR. I have already backported transcalc. See my comments here #142 But rescodes utility was removed from Qucs distribution and should not be reverted.
* I have seen that you have deleted all Ngspice and Xyce examples
* The LaTeX documentation in `qucs-doc` subdirectory is related to Qucs project.  The usage of Qucs and Qucs-S may be different. See also #78 Qucs documentation should not be distributed with Qucs-S because it can mislead user. Help menu points to Qucs-S online documentation now. 
* Some files have 755 permissions instead of 644

I may consider your PR again after you fix the mentioned issue and split it into individual smaller PRs.

Issue #162, opened 2022-09-20
Make revision of the examples


ra3xdh:
The examples from `examples` directory were made for Qucsator simulation kernel and may be incompatible with Qucs-S+Ngspice. Like `examples/resonance.sch` will not work work with Ngspice and produce cryptic error message. This may be misleading for new users. It's need to review all examples and move inconpatible with Ngspice to special `qucsator` subdirectory.   Another task is to add more examples.

- [x] Move Ngspice-incompatible examples elsewhere
- [x] Add new examples
- [x] Revise existing example schematic. Something may become outdated there

tomhajjar:
I've segregated the examples into separate "schematic" folders under separate start folders for Qucs, Qucs-s and QucsStudio. "prj" folders are also segregated by qucs version. I have tested almost every schematic and project.  

![2022-09-20_222120](https://user-images.githubusercontent.com/102838604/191400018-5bdca874-526b-467b-b57f-703adb00e935.jpg)

tomhajjar:
One of the issues I see with the examples is the non-descriptive names used for schematics. Unless there is a directory, no user is going to know what many of the circuits are. 

When I look for help trying to do something, I often use Grep looking for keywords in the examples folder. I then look at the example schematic hoping it will help. Since most of the examples are for "Qucs", it isn't as helpful as it could be.

tomhajjar:
When "porting" Qucs schematics to Qucs-S, I'm slowed down by Qucs-S not generating any error messages. I don't know ngspice well enough to convert the two Parameter Sweeps used in Qucs to work for Qucs-S. Getting no error messages doesn't help.

![2022-09-21_125106](https://user-images.githubusercontent.com/102838604/191566157-a7296776-3d3d-4e76-b6e9-bfeed3b9009b.jpg)
![2022-09-21_125753](https://user-images.githubusercontent.com/102838604/191566174-95e14cf3-adaf-4e33-84f6-fc7eeac3d425.jpg)

[schematics-S.zip](https://github.com/ra3xdh/qucs_s/files/9618872/schematics-S.zip)

tomhajjar:
There are a number of examples that use BJT with Substrate. Qucs-S doesn't support this device but I believe ngspice and xyce does.

tomhajjar:
First cut at the revised examples. ngspice and xyce folders to follow

[schematics-S.zip](https://github.com/ra3xdh/qucs_s/files/9621906/schematics-S.zip)

[...]

PR #168, opened 2022-10-01
simple splitting to allow multiple parameters in a single sweep

KAGVa:
Splitting the Params string with ';' (hardcoded as protected QString for Param_Sweep class atm, might be useful to be able to change that later?) and adding for each splitted element a alter XX =  $element[0]_act in the netlist

tested with ubuntu 22.04 and works (creates a correct netlist, ngspice3.7 does not complain and the result I tested looks right)


ra3xdh:
Thanks for the contribution! I was considering to introduce a similar feature with splitting device list. This will resolve parameter sweep issue reported in #162. But you need to retarget this PR to `current` branch to proceed. The `master` branch represents the stable state of the package, but `current` branch represents development state.

KAGVa:
closed this pull request and reopened a new one for current branch -> it is now PR169

PR #170, opened 2022-10-06
Add '+' to start of multiline model continuations

cactorium:
I'm not sure whether this is expected behavior, but when you add in a multiline model you need to manually add in a '+' to each additional line. This adds it for you, no clue whether this is better or worse than the default behavior

### DIFF
--- a/qucs/spicecomponents/sp_model.cpp
+++ b/qucs/spicecomponents/sp_model.cpp
@@ -80,9 +80,17 @@ QString S4Q_Model::getSpiceModel()
 
     QString s;
     s.clear();
+
+    bool first_line = true;
     foreach (Property *pp, Props) {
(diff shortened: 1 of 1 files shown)


ra3xdh:
Hello Kelvin, the automatic addition of the `+` sign will break the backward compatibility of the existing schematics. Also Qucs-S supports drag'n*drop of the `.MODEL` from the SPICE libraries if it is copied to the `$HOME/user_lib`. See https://qucs-s-help.readthedocs.io/en/latest/SubLib.html#usage-of-the-whole-spice-library for more info. The `+` sign in the beginning of the line is copied automatically in such case. So, I cannot accept this pull request.

cactorium:
That's fair, thank you for the explanation

Issue #171, opened 2022-10-07
"current" branch will not build, missing "qucslib_common.h"


tvrusso:
I just tried to build the code off of the current branch today and it failed.  qucs/qucs.cpp includes "qucslib_common.h" but that file no longer exists in the source tree.  I tried just removing the include, but there were undeclared identifiers and other issues, and other files also include qucslib_common.h.

ra3xdh:
Hello Tom, I forgot to commit `qucslib_common.h` after removing the `qucs-lib` tool #139. I have just restored the missing file. Please try to compile again. I have recently added CMake build test on the `master` branch to prevent such issues in the future https://github.com/ra3xdh/qucs_s/actions/workflows/cmake.yml This will be available for all branches after merging `master` and `current`

tvrusso:
Thanks, Vadim.  Yes, it is fine now.

I had just created a pull request while you were fixing it, but you beat me to the punch.

PR #172, opened 2022-10-07
Restore qucslib_common.h file

tvrusso:
It appears that when qucs-lib was merged, it was intended that the file qucslib_common.h would have been moved to the qucs subdirectory.  That move appears to have been forgotten, or perhaps just not committed.

This pull request restores the file by copying it out of the master branch.

If accepted, it would close issue #171.


tvrusso:
Ooops.  I see that as I entered this pull request, you have already restored the file.

Closing.

Issue #133, opened 2022-08-11
xyce examples don't work


tomhajjar:
I found a number of the supplied xyce Examples that don't work. Attached are a few I fixed. 

Does xyce have the unwrap phase function available? Ultimately I want to be able to generate Group Delay for S-parameters. 

[2N2222_Amplifiers_xyce_prj.zip](https://github.com/ra3xdh/qucs_s/files/9312097/2N2222_Amplifiers_xyce_prj.zip)

tomhajjar:
I found two schematics used R1=39K Ohm instead of 39k Ohm. 

Another issue caused xyce simulation to be wrong and Qucs-S to randomly crash.

R5=1M Ohm worked
R5=1MEG  worked
R5=1M bad news

tomhajjar:
[Temp2.zip](https://github.com/ra3xdh/qucs_s/files/9313198/Temp2.zip)

ra3xdh:
>  the supplied xyce Examples that don't work

Could you point exactly what doesn't work? The schematics from attached project are simulated without error except the issue with units (see my explanations below).

> Does xyce have the unwrap phase function available?

Xyce has very poor implementation of the postprocessor functions compared with Ngspice. I cannot find something like `cph()` in the Xyce manual. You may achieve complex measurement like the group delay only with export of the dataset and processing it with external application. 

> Another issue caused xyce simulation to be wrong and Qucs-S to randomly crash.

I have tried different combinations of the resitance but I was not able to reproduce application crash. The `1 M` value is treated as 1mOhm because the SPICE is registry independent. That's why you obtain short circuit on output and nearly zero output voltage. 

The Qucsator was registry dependent unlike SPICE and used different unit suffixes for units. The logic of units in Qucs-S is the following:

* If no unit is specified like (1k 10m) assume registry-independent SPICE notation. The 1M will 1m=1e-3 but not 1e6. The 1Meg will 1e6
* If unit (Ohm, V, F, etc) is specified assume Qucsator notation like 1kOhm = 1000 Ohm 1MOhm = 1000000 Ohm.

tomhajjar:
I sent you the xyce examples after they were corrected and cleaned up. The two "bad" schematics used R1=39K Ohm instead of 39k Ohm. Attached are the original file folders and the entire test suite. I forgot where I downloaded them from. 

Qucs crashed a coupe of times when R5=1M. I cannot say what additional actions are needed to trigger the crash.

[...]

Issue #76, opened 2022-02-22
Qucs-S crashes on Wayland


ra3xdh:
The crash on Wayland reported by one of the users. The following command sequence is used to launch Qucs-S: 
~~~
$ QT_QPA_PLATFORM=wayland
$ qucs-s
Info Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
~~~

After that Qucs-S crashes. I can reproduce this bug using the Qucs-S RPM package with the CentOS_Stream in VirtualBox that using Wayland by default. The `qtwayland` package is installed. I would appreciate help with this issue. I guess that some Qt configuration options must be set to run application without crash.

ra3xdh:
Fixed by https://github.com/ra3xdh/qucs_s/commit/2dc8cfbc33284f7abb87f4f3792f23ad4e76cbc8

PR #201, opened 2022-12-29
Fix negative screen coordinates

zergud:
- related to #133 (win64 math tolerance)
- rewrite MessageHandler to more verbose console output and correct encoding

### DIFF
--- a/qucs/diagrams/graph.cpp
+++ b/qucs/diagrams/graph.cpp
@@ -16,8 +16,9 @@
  ***************************************************************************/
 #include "graph.h"
 
-#include <stdlib.h>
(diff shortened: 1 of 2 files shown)


ra3xdh:
Thanks for the fix! I confirm that the auto-scaling issue reported in #133 has gone after applying this patch. I will merge this after I check the new message handler operation on the Linux platform using Wayland. In the past a crash related to message handler was observed on Fedora Linux distribution #76

ra3xdh:
I have tested the new MessageHandler with the latest Fedora and found no problems. Merging this. Thanks for the contribution!

PR #215, opened 2023-02-13
qt6 native win build fixes

dsm:
fixed namespace std usage made win10 build break. I changed cmake options for compiler agnostic version. I didn't made it work for msvc build but mingw qt6.5.0 build was succesfully done in native win 10 env.

Ekran görüntüsü 2023-02-13 135054


ra3xdh:
Thanks for the contribution! I have merged this. I have also added the setting of `-fPIC` compiler flag using `SET(CMAKE_POSITION_INDEPENDENT_CODE ON)`. Otherwise build fails for Linux.

PR #217, opened 2023-02-13
updates for related to modernization of cmake

dsm:
Hi, no need for WITH_QT6 cmake define, qt version find changed, I tested windows but not tested others.


dsm:
this pr will break build with qt version less than 5.15

qucs/CMakeLists.txt
```diff
@@ -266,7 +266,7 @@ ADD_EXECUTABLE( ${QUCS_NAME} MACOSX_BUNDLE WIN32
 # Tell CMake which libraries we need to link our executable against.
 #
 TARGET_LINK_LIBRARIES( ${QUCS_NAME}  components diagrams dialogs paintings extsimkernels spicecomponents qt3_compat ${QT_LIBRARIES} )
-SET_TARGET_PROPERTIES(${QUCS_NAME} PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
```
ra3xdh, line 269: The duplicate option here is done intentionally. The Qucs application may be built separately from the utilities for debug purposes. If toplevel Cmake is not accessible the build will fail without setting PIC option here.

qucs-activefilter/CMakeLists.txt
```diff
@@ -44,30 +28,31 @@ CONFIGURE_FILE (
...
 
-IF(WITH_QT6)
-	FIND_PACKAGE( Qt6 COMPONENTS Core Gui Widgets Svg SvgWidgets REQUIRED)
-
-INCLUDE_DIRECTORIES(
-      ${Qt6Core_INCLUDE_DIRS}
-      ${Qt6Widgets_INCLUDE_DIRS}
-      ${Qt6Svg_INCLUDE_DIRS}
-      ${Qt6SvgWidgets_INCLUDE_DIRS}
+find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Core Gui Widgets Svg SvgWidgets)
```
ra3xdh, line 31: Linux build with Qt5 fails on this line. Qt5 doesn't provide SvgWidgets package, But CMake tries to find it. Using CMake 3.18 on Ubuntu18.04.

dsm: I think latest commit will solve this.

ra3xdh: Yes, the problem with SvgWidgets library has been resolved.

qucs/CMakeLists.txt
ra3xdh: Yes, the problem with PIC flag has been resolved too.

ra3xdh:
I have tested a build system update on Ubuntu22.04 and 18.04 and found no major issue. Will you add more commits to this PR?

One question regarding automatic Qt version check. Is it possible to specify the Qt version manually using a new build system? This may be needed for debug purposes. I have tested this on Ubunutu22.04 having both Qt5 and Qt6 installed and build system always finds Qt5 automatically.

dsm:
I will add default qt version variable

ra3xdh:
Waiting for Qt version select options and then I will merge this PR.

dsm:
qt doc said find_package search first qt6 if fails then search qt5 why this happen, I don't now but I set QT_DEFAULT_MAJOR_VERSION to 6 maybe this will help to solve this can you test it ?

dsm:
if we write good deployment using github actions and release process and pr merge all version an all supported os. current one is very poor. I'm not experienced but I will help to develop new one.

ra3xdh:
Default Qt version should be Qt5. It's need to provide a method to manually select the desired Qt version.

dsm:
what is your QT_DIR or QT env variable. this variable set your qt5 dir this will find qt5 can you check this

dsm:
this command will find qt6 first
`cmake -DCMAKE_PREFIX_PATH="C:\Qt\6.5.0\mingw_64;C:\Qt\5.15.2\mingw81_64" -G "Unix Makefiles" ..`

this command will find qt5 first
`cmake -DCMAKE_PREFIX_PATH="C:\Qt\5.15.2\mingw81_64;C:\Qt\6.5.0\mingw_64" -G "Unix Makefiles" ..`

or set QT_DIR env variable to which one will select. Current WITH_QT6=ON doesn't solve this too.

ra3xdh:
I have tried different combination on Ubuntu22.04 I seems that `QT_PATH` variable has no effect for Cmake on Linux platform.

ra3xdh:
> if we write good deployment using github actions 

I am using OpenSUSE build service system (OBS) to build DEB and RPM packages for Linux distributions. The Arch has its own packaging system and generates package for AUR independently. The Windows package is generated manually using MXE cross-compiler. As far I know it is not possible to automatize such package generation procedure using github actions.

dsm:
external CI also possible but so many open source repo use github ci to release linux mac freebsd and windows build. I think this is possible but take some time

dsm:
[example workflow](https://github.com/Serial-Studio/Serial-Studio/blob/master/.github/workflows/deploy.yml) - in this repo define a qt version inside a yml file and all os build upload as continuous build release

dsm:
I will add msvc support

ra3xdh:
> I have tried different combination on Ubuntu22.04 It seems that QT_DIR variable has no effect for Cmake on Linux 

More details on this. I have tried the following cmake varaibles combinations combinations. All of them have no effect on Ubuntu22.04 with both Qt5 and Qt6 installed from repository. The Qt6 installation directory is `/usr/lib/qt6` 

* The `QT_DIR` setting has no effect
~~~
export QT_DIR=/usr/lib/qt6
cmake ..
~~~
* The `CMAKE_PREFIX_PATH` also has no effect
~~~
cmake -DCMAKE_PREFIX_PATH=/usr/lib/qt6 ..
~~~

 Maybe I am doing something wrong. I have also tried clearing the CMake cache before the CMake execution. It remains unclear how to specify the Qt version for CMake explicitly.  

The summary:

* The default Qt version must be Qt5
* It's need to provide a possibility to switch build to Qt6 manually 

After these two points will be resolved, I will merge this PR.

dsm:
I will resolve this issue and on my current machine installed on fedora 37

dsm:
for qt5 -> export QT_DIR=/usr/lib/x86_64-linux-gnu/cmake/Qt5
for qt6 -> export QT_DIR=/usr/lib/x86_64-linux-gnu/cmake/Qt6

clear cmake cache and build dir &&  cmake .. this will work I tried ubuntu 22.04

ra3xdh:
Yes, the environment variable setting in a such way works as expected. The PR could be merged now.

ra3xdh:
I will update build instructions in the README after the merge.

PR #238, opened 2023-03-04
Fix for voltage and current sources

Radvall:
Added space between the node and the parameters for the non-linear voltage and current sources;
Added phase for the ideal ac voltage and current sources.


ra3xdh:
Thanks for contribution! Yes, the ACPHASE was indeed missing in the AC sources.

PR #243, opened 2023-03-11
update python code for post-analysis

Radvall:
Moved parse_qucs_result function to QucsDataset class method and improved work with independent variables.
Add extra example to show work with independent variables when schematic have 2 different simulations.

The results of running the python scripts are shown below. I use Python 3.10.9 and numpy 1.23.0 for tests.

parse_result_example2.py;
parse_result_example2

parse_result_example.py:
parse_result_example


ra3xdh:
Thanks for the contribution! Python scripts for dataset processing was not updated in the recent years. I have merged this.

PR #245, opened 2023-03-13
some improvements.

dsm:
this pr really small but I made some improvements convert c array to std::array make more readable and fixed unnecessary memory usage of char Mnemonic[16] ...


ra3xdh:
Thanks for the contribution! I have added explicit `array` header in `mnemo.cpp` and merged this PR.

PR #262, opened 2023-04-20
Added spice entry for flip-flops

Radvall:
I added the spice entry to the existing analog JK flip-flop /w SR, D flip-flop /w SR, T flip-flop flip-flops.
The simulation results are shown below and the circuit is added to the examples/ngspice dirrectory. 

flip_flops

From the digital components I added the D-flip-flop. The unused output QB was connected to ground through a 1N capacitor. 
And the Set output was connected directly to ground.


ra3xdh:
I have tested new devices and found no visible issues. Merging this. Thanks for the contribution!

PR #265, opened 2023-04-26
Added digital source

Radvall:
Added a new method convert_notation to the spicecompat class that replaces common unit suffixes and SI prefixes in a value string with their corresponding exponent values. The supported unit suffixes include "p", "n", "u", "m", "k", "Meg", "G", and "T".
Added spice entry for digital source.

The digital source was created as a VPWL source, with the rise and fall times set to one hundredth of the first time step value. The main difference between the original digital source and this implementation is that the signal is not periodic.

The comparison between the two versions is shown below.
ngspice_vs_qucsator

### DIFF
--- a/qucs/components/digi_source.cpp
+++ b/qucs/components/digi_source.cpp
@@ -17,7 +17,8 @@
 #include "digi_source.h"
 #include "node.h"
 #include "misc.h"
-
+#include "extsimkernels/spicecompat.h"
+#include <QDebug>
 
 Digi_Source::Digi_Source()
 {
@@ -46,6 +47,7 @@ Digi_Source::Digi_Source()
   ty = y2+2;
   Model = "DigiSource";
   Name  = "S";
+  SpiceModel = "V";
 
   // This property must stay in this order !
   Props.append(new Property("Num", "1", true,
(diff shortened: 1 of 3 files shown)


qucs/extsimkernels/spicecompat.cpp
```diff
@@ -74,6 +75,41 @@ QString spicecompat::normalize_value(QString Value)
...
+ *        prefixes in a string with their corresponding exponent values.
+ *        For example, "1.23u" would be converted to "1.23E-06",
+ *        and "1.23m" would be converted to "1.23E-03".
+ *
+ *        Supported unit suffixes include "p", "n", "u", "m", "k", "Meg", "G", and "T".
+ *
+ * \param Value The input string to convert.
+ * \return The string with converted notation.
+ */
+QString spicecompat::convert_notation(QString Value)
```
ra3xdh, line 89: This implementation may not work properly if called on variable but not a number with a suffix. It will replace all occurence of suffix (for example "P" letter). The proper implementation could be done with regexp.

qucs/components/digi_source.cpp
```diff
@@ -186,3 +188,83 @@ QString Digi_Source::verilogCode(int NumPorts)
...
   return s;
 }
+
+QString Digi_Source::spice_netlist(bool)
+{
+  QString s    = SpiceModel + Name;
+  QString port = spicecompat::normalize_node_name(Ports.at(0)->Connection->Name);
+  s += " " + port + " 0 "; // node names
+
+  QString V    = spicecompat::normalize_value(getProperty("V")->Value);
```
ra3xdh, line 198: It's need to check if PWL source allows using variables inside PWL function. If variables are not allowed, it would be sufficient to simple use `misc::str2num`

ra3xdh:
Thanks for the contribution! But there are some potential problems with new `spicecompat::convert_notation`. It's need to define if we need variables as digital sources parameters.

ra3xdh:
Also why `spicecompat::convert_notation` was needed? Are there problems with PWL to accept the suffixes?

qucs/components/digi_source.cpp
Radvall: PWL source allows using variables inside PWL function. But in this implementation of digital source it can not work due to the style of writing times in original QUCS digital source.

Radvall:
> Also why `spicecompat::convert_notation` was needed? Are there problems with PWL to accept the suffixes?

No, there aren't any problems with PWL accepting suffixes. But I can't use times from the digital source directly in the PWL source due to the style of writing times in original QUCS digital source. That is why I wrote `spicecompat::convert_notation` .

For example, if we specify times as "1n;2n" for the digital source with an initial output value of low and a high-level voltage of 1V 
this means that the signal will start at 0V and at time 1ns changed to 1V, and end at time t=3ns.
The PWL signal for this example is defined by 6 time and value pairs. The pairs are (0, 0), (1E-09, 0), (1.01E-09, 1), (3E-09, 1), (3.01E-09, 0), and (3.02E-09, 0). Where 1E-11 is the rise or fall time as 0.01 of the first time step.

The first pair (0, 0) specifies that the signal starts at 0 volts at time t=0. The next pair (1E-09, 0) specifies that the signal remains at 0 volts until time t=1E-09.
The third pair (1.01E-09, 1) specifies that the signal rises linearly from 0 volts to 1 volt over a period of 1 picosecond, which is the rise time.
The signal remains at 1 volt between time t=1.01E-09 and t=3E-09, which is the specified duration of 1n+2n.
The fifth pair (3.01E-09, 0) specifies that the signal falls linearly from 1 volt to 0 volts over a period of 1 picosecond, which is the fall time.
The signal remains at 0 volts from time t=3.01E-09 until the end of the specified time range at t=2n. It made to set the last time step to 0 regardless of initial outputs values.

ra3xdh:
Thanks for explanation. My advice is to use the `misc::str2num` instead of `spicecompat::convert_notation` and then add numbers.

Radvall:
I have replaced spicecompat::convert_notation with misc::str2num.
But I see some problems with using misc::str2num for SPICE netlist. The first is case sensitivity, the second is incorrect handling of the 1e6 prefix.
For NGSPICE it will be Meg instead of M as used in misc::str2num. However, for times in a digital source this should not cause problems.

ra3xdh:
I have merged this PR into the `current`. Thanks for the contribution!

Issue #251, opened 2023-03-25
Trying digital components


0range0range:
Hello, 

I was trying to play around with some digital components and I could not get it to work at first. 
I had an error: 
```
ERROR - Output delay <= 0 not allowed - output ignored!
  Instance: ay1
  Node: out1
  Time: 0.001000 
```

After reading: https://github.com/ra3xdh/qucs_s/issues/97 , I noticed a `t` parameter for digital blocks and setting it to `1ns` was enough to overcome the error.

1. But now I wonder why does it work and what is the role of `t`? 

2. Wouldn't it be better to set it to 1n by default (as for example for rise and fall time for square wave source) so new people do not have to deal with it when they just want to try 1-2 gates?

3. I tried to change `Voltage of high level` parameter, but it always outputs 3.3V. How does this one works?

Here is the circuit example
![image](https://user-images.githubusercontent.com/95714694/227711864-7829f094-9dab-407d-9231-01f1086f1658.png)

**TODO** list:
 - [x] Set the default delay value to 1ns instead of 0. Otherwise it will cause XSPICE error.

ra3xdh:
> But now I wonder why does it work and what is the role of t?

The `t` represents the delay. See here for more information https://ngspice.sourceforge.io/docs/ngspice-html-manual/manual.xhtml#magicparlabel-89763 Existing Qucs symbol doesn't have the separate property for rise and fall delay and assumes t=tr=tf. Otherwise it would require the redesign of the symbol and the loss of the backward compatibility. 

> Wouldn't it be better to set it to 1n by default

Yes, this could be done. The XSPICE models don't support zero delay value. 

> I tried to change Voltage of high level parameter, but it always outputs 3.3V.

This property has no effects with XSPICE devices. You need to define the `vcc` parameter that is used as the VCC voltage of the logic gate devices.
![image](https://user-images.githubusercontent.com/4920080/227712999-a366e059-04c2-47e1-a664-f0cc909625e7.png)

ra3xdh:
I have changed default `t` value to prevent errors coming from XSPICE.

tomhajjar:
None of the digital examples in Qucs can be used in Qucs-S. 

Will additional digital components and sources available in Qucs be available in Qucs-S?

ra3xdh:
[...]

Issue #97, opened 2022-07-03
Bring digital simulation back to Qucs-S


ra3xdh:
The Ngspice team is currently developing the support for U devices from PSPICE representing digital blocks. This feature is expected in Ngspice-38 release. The support for U devices makes possible to bring the digital components back to Qucs-S. Currently digital devices are hidden as SPICE-incompatible. 
Holger @holvo , please let me know, when the U-devices will be ready to start its implementation in Qucs-S GUI. 

**TODO** list in order to revert digital simulation for Qucs-S

- [x] Add XSPICE logic gates models
- [x] Add XSPICE flip-flop models
- [x] Add SPICE entries for digital sources in analog mode
- [x] Revert optional fully digital simulation mode using `iverilog`
- [x] Backport `qucsveri` script from the Qucs-0.0.20-rc1. This script is needed to start the `iverilog`
- [x] Backport `qucsconv` utility that is needed to convert VCD files to Qucs XML dataset
- [x] Revert hidden digital devices to components panel
- [x] Fix remaining issues with default delay time and show netlist mentioned in #272 
- [x] Add tutorial section on digital simulation; old Qucs tutorial is not valid anymore

ra3xdh:
I have started the implementation of the digital devices based on the recent Ngspice-38. Two models are back: ideal inverted and NAND gate. The manual build of the Ngspice from `pre-master` branch is required. Everything seems to work as expected. I have pushed the code to the `digital` branch https://github.com/ra3xdh/qucs_s/tree/digital The attached screenshot illustrates the operation of the RC-generator based on logical gates.
![image](https://user-images.githubusercontent.com/4920080/193865600-e927d1c9-24c1-4920-8843-f87239e3de88.png)

ra3xdh:
I have added a TODO list for this issue.

ra3xdh:
SPICE entry for digital source in analog mode was added by #265

ra3xdh:
I have failed to make a full backport of the `qucsconv` tool, because it depends on almost half of Qucs-S code. I had to restrict only to VCD files support and renamed the utility to `vcd2qucsdat`. I also renamed the wrapper scripts to avoid name conflict with old Qucs package. The new names for the wrapper scripts are:

~~~
qucsdigilib => qucs_mkdigilib
qucsdigi => qucs_run_hdl
qucsveri => qucs_run_verilog
~~~

[...]

PR #272, opened 2023-05-24
Bring back digital modeling

Radvall:
Finished work on bringing back digital modeling using iverilog and freeHDL. I checked the work on a simple circuit for VHDL and verilog models.  Simulation results are recorded in *.dat format (qucsator dataset format).

and_gate_verilog_model
and_gate_VHDL_model


ra3xdh:
Thanks for the contribution! I will check its operation and report feedback in the next few days.

ra3xdh:
I have noticed the following problems while testing the PR:

* After pressing F6 key the application shows the netlist `spice4qucs.cir` instead of `netlist.txt` that is valid for digital simulation
* The Verilog netlister doesn't accept default t=1e-9 delay for logic gates if scientific notation is used. It requires t=1ns. See also #251 

The mentioned issues could be easily fixed. So I am merging this PR in to the `current` despite the remaining issues.

ra3xdh:
I have merged this. See updated TODO list of the #97 for remaining issues with digital simulation.

PR #276, opened 2023-06-02
New SelectionRect

zergud:
**fixes:**
- selection rectangle sometimes did not disappear after mouseup   

**new:**
- new semitrasparent selection rectangle
- new selection option: with SHIFT pressed we can select components partially intersected by selection rect


ra3xdh:
I have merged this into the `current` branch. New selection rectangle look more modern and works well. Thanks for the contribution!

PR #283, opened 2023-06-15
Redesign simulator switching system

### LINKED ISSUE #274: Redesign of simulator switching system
Current simulator switching system has some limitations and would benefit from redesign. The simulator switching system should meet the following requirements:

* It should be possible to switch simulator without requirement to restart the application
* Place a "simulator" drop-down list someweher on visible place in the GUI
* This drop-down list should contain only properly configured simulators which binaries are found in the system. The user cannot switch to invalid simulator by mistake
* After the simulator is switched, the incompatible devices become hidden. The application restart will be not needed anymore. If the incompatible devices are already added on the schematic, it should be marked by some way. I propose black cross for this purpose. Red cross is for disabled devices, green gross for shorted devices.
* The simulator binaries location paths could be configured from **Simulation->Select simulator** menu.

zergud:
fixes #274


ra3xdh:
I have tested this PR and found the following problems:

* New system cannot recognize simulator if it is defined using `$PATH`. This case is common for Linux platform. For example it cannot find qucsator on my Linux system.
image

I suggest to implement a custom search procedure instead of using simple `QFile::exists()`. It should look for executable also in `$PATH`.

* An addition to the previous point. I have implemented an automatic search of the Ngspice exexcutable in order to prepare application for the build of Windows installer. This feature may be combined to simulator selection. See https://github.com/ra3xdh/qucs_s/commit/7a77ceb3a58efcb42a53dc7f6f350a96a3fc2f0e 

* It is possible to set simulator to `Not specified` and simulation doesn't work in such case. Only simulators should be available in the drop-down list.
image

* On the first start the applications reads old settings from config file and sets simulator to `Not specified`. As the solution I may suggest to rename the setting in the config (see `saveSettings()` and `loadSettings` in `main.cpp`). Also it may help to shift bit mask of the `spicecompat::Simulator` The range from 0x00 to 0x0F will indicate old settings before v1.1.1 and it's need to show the settings dialog on the first start again. 

* The blue BJT, MOSFET, and JFET symbols should be available for all simulators, but not only for Qucsator.

ra3xdh:
Now everything seems to work as expected except the problem with simulators in `$PATH` I will try to provide the solution for this problem in the next few days.

ra3xdh:
I have tested it again and found no remaining problems. Merging this. Thanks for the contribution!

PR #285, opened 2023-06-17
fix print to pdf on windows

### LINKED ISSUE #284: Possible issue, using MS windows pdf printer
Hi, when printing with the MS windows pdf printer I can't get the text to stay in proportion to what I set in the .sch

print_pdf_3

zergud:
fixes #284


ra3xdh:
Thanks for the quick fix!

PR #331, opened 2023-10-18
simsettingsdialog.cpp: increase width window size 500 -> 750

antohami:
There is not enough space for Russian translation.


ra3xdh:
Thanks for reporting this, but I decided to fix this with another way. I have corrected translation so the length or source and translated strings is matched. See https://github.com/ra3xdh/qucs_s/commit/fddc9f0384518bc0b074748b06cdc9b38eceb6ce Closing without merge.

Issue #353, opened 2023-11-04
Switch time controlled


ra3xdh:
### Discussed in https://github.com/ra3xdh/qucs_s/discussions/348

<div type='discussions-op-text'>

<sup>Originally posted by **cmat21** November  1, 2023</sup>
Hi, is there any possibility to add a "switch time controlled" component to ngspice?
![Switch_time_controlled](https://github.com/ra3xdh/qucs_s/assets/70080894/06dd839f-da68-4e72-ac49-b44321f039f9)
</div>

ra3xdh:
Switch implementation for SPICE added by #355. Closing.

PR #355, opened 2023-11-09
Added spice entry for time-controlled switch

Radvall:
I've made some enhancements to the spice model:
- Added a spice entry for time-controlled switch #353;
- Updated descriptions for "Transition" and "Temp" parameters;
- Updated the default value for Ron.

The simulation test results are shown bellow:
switch_test


ra3xdh:
Thanks for contribution!

PR #397, opened 2023-12-11
Rewrite rotation of paintings

### LINKED ISSUE #203: Symbol lines shifting off "grid" when rotated
When creating symbols for components, if you cut or copy a symbol AND rotate the symbol while pasting, the lines drawn on an angle will shift off the "grid". Attached shows shifting caused by rotate left x2. Mirroring a symbol doesn't seem to cause issues.

Attached is an update on the symbols.sch I created. It is very useful when making new symbols.

In the past I asked if Rotate Right could be added. It can save a lot of time as well as minimize shifting off grid.

[schematic_symbols.zip](https://github.com/ra3xdh/qucs_s/files/10355342/schematic_symbols.zip)

zergud:
fixes #203

### DIFF
--- a/qucs/schematic.cpp
+++ b/qucs/schematic.cpp
@@ -15,626 +15,641 @@
  *                                                                         *
  ***************************************************************************/
 
-#include <stdlib.h>
 #include <limits.h>
+#include <stdlib.h>
 
-#include <QFileInfo>
-#include <QPrinter>
-#include <QPaintDevice>
-#include <QDir>
-#include <QTextStream>
-#include <QDragLeaveEvent>
-#include <qt3_compat/qt_compat.h>
-#include <QPixmap>
-#include <QDragEnterEvent>
-#include <QDragMoveEvent>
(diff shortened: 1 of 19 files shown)


qucs/schematic.cpp
```diff
@@ -15,626 +15,641 @@
...
+Q3PtrList<Node> SymbolNodes;
+Q3PtrList<Diagram> SymbolDiags;
 Q3PtrList<Component> SymbolComps;
 
-
-Schematic::Schematic(QucsApp *App_, const QString& Name_)
+Schematic::Schematic(QucsApp *App_, const QString &Name_)
     : QucsDoc(App_, Name_)
 {
-  symbolMode = false;
```
ra3xdh, line 69: Indentation should be preserved here. Otherwise git shows 100% code replacement, which make hard to track the changes.

ra3xdh:
Yes, the issue with rotation has been fixed. I didn't find side effects. Merging.

PR #412, opened 2023-12-27
Add support for custom prefix to dataset variables

ivandi69:
The prefix is enclosed between # # in the output file name. If the file name contains a #custom_prefix# section it will be prepended to the dataset variables. Useful for custom simulations where one can set multiple simulations and multiple output files.

Example nutmeg script:
```
AC DEC 10 1MEG 1G
write custom#ac1#.txt v(out)
destroy all

AC LIN 100 90MEG 110MEG
write custom#ac2#.txt v(out)
destroy all
```
The dataset will contain:
ac1.ac.v(out) ac1.frequency
ac2.ac.v(out) ac2.frequency

Prevent duplications in simulations QStringList of Ngspice::createNetlist. With current code if one has two nutmeg scripts both scripts are placed twice in the control section of the netlist.

### DIFF
--- a/qucs/extsimkernels/abstractspicekernel.cpp
+++ b/qucs/extsimkernels/abstractspicekernel.cpp
@@ -1107,6 +1107,8 @@ void AbstractSpiceKernel::convertToQucsData(const QString &qucs_dataset)
         bool hasParSweep = false;
         bool hasDblParSweep = false;
 
+        QRegularExpression custom_prefix_rx("(?<=#).*?(?=#)");
+        QString custom_prefix = custom_prefix_rx.match(ngspice_output_filename).captured(0);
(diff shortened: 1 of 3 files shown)


ra3xdh:
Thanks for the contribution! Merged. In the long perspective this feature should be documented somewhere, otherwise it will be lost. Unfortunately I only started to write a new documentation for Qucs-S and there is no place where you can propose a PR with the documentation addition.

ivandi69:
A note could be added to the default text of the "Edit SPICE Code" dialog.
-------------

AC LIN 2000 100 10MEG

let K=V(1)/V(2)


* Extra output

* A custom prefix could be placed between # #

* It will be prepended to all dataset variables

write custom#ac1#.txt K

--------------

Thanks

PR #420, opened 2023-12-29
Add "zoom to pointer" behaviour for zooming using mouse wheel

wawuwo:
Hello! I've found out that it is quite hard to explain this zooming feature in plain words, so I'm attaching "before" and "after" videos. Nonetheless, let me try to explain what this PR is about. 

This PR alters "zoom in/out using mouse wheel" behavior so that the part of contents under the mouse pointer doesn't change its position relative to app's window while contents is being scaled. In effect it looks like you're "zooming into" the mouse pointer or "zooming out of" the mouse pointer.

PR is of two commits: one adds the described functionality, another adds a couple of comments which I think may be useful in future (it took some time for me to understand the "zooming" and "rendering" code, so I think it would be nice to leave some hints in source for others).
[after.webm](https://github.com/ra3xdh/qucs_s/assets/40355531/543c53a2-425c-4bd6-9898-f2b94077b64f)
[before.webm](https://github.com/ra3xdh/qucs_s/assets/40355531/756d8256-8cea-44ce-b191-8638aa0e1c4b)

### DIFF
--- a/qucs/schematic.cpp
+++ b/qucs/schematic.cpp
@@ -35,7 +35,9 @@
 #include <QPaintDevice>
 #include <QPainter>
 #include <QPixmap>
+#include <QPoint>
 #include <QPrinter>
+#include <QRect>
 #include <QTextStream>
 #include <QUrl>
 #include <QWheelEvent>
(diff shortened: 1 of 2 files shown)


zergud:
Good job!
it would be cool to implement PAN by mouse in future ;)

wawuwo:
@zergud Thanks! I might try to look into it one day, but unfortunately I can't give any promises

ra3xdh:
Thanks for the contribution! Merging.

iolpltuisciaoln:
Really can't pan with mouse its something new

wawuwo:
@zergud @iolpltuisciaoln #488 introduces panning with mouse

PR #431, opened 2024-01-03
Modify SPICE RCD devices; fixes #349

### LINKED ISSUE #349: Modify SPICE RCLDQ devices to provide support for microelectronics PDKs
I have recently added unified MOS devices that could be mapped to X letter. See #128 and #242. Other primitive devices also need to be modified to ensure compatibility with PDKs. It's need to add possibility to define pins number and letter. The following improvements should be done:

- [ ] Add terminal number and letter (C,X) property for capacitor  (needed 3-terminal capacitor)
- [ ] Add terminal number and letter (R,X) property for resistor (needed 3-terminal resistor)
- [ ] Define letter (D,X) and pins count for diode (needed 3-terminal diode)
- [x] Add terminal number and letter (Q,X,N) property for BJT

Radvall:
I added X letter and pin-number proprieties to the existing R_SPICE, C_SPICE and DIODE_SPICE devices, placing them at the end of the proprieties list. This should not affect compatibility with older schematic versions. The new symbols are shown bellow:

new_symbols_qucs

I turned off rotation for R_SPICE and C_SPICE components as this fix the selection box during editing when clicking apply and when inserting a component after CTR+C CTR+V.

selection_box_res

Also I added 3-pin symbols to the left panel but I'm not sure if they are needed there.

### DIFF
--- a/qucs/module.cpp
+++ b/qucs/module.cpp
@@ -180,6 +180,8 @@ void Module::intoCategory (Module * m) {
 
 #define REGISTER_COMP_1(cat,val) \
   registerComponent (cat, &val::info)
+#define REGISTER_SPECIFIC_COMP_1(cat,val,inf1) \
+registerComponent (cat, &val::inf1)
 #define REGISTER_COMP_2(cat,val,inf1,inf2) \
   registerComponent (cat, &val::inf1); \
   registerComponent (cat, &val::inf2)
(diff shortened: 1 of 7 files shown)


ra3xdh:
Looks impressive, thanks for the contribution! There is a typo in DIODE_SPICE.cpp See my line comment.

ra3xdh:
> Also I added 3-pin symbols to the left panel but I'm not sure if they are needed there.

It makes sense to add a new group named "PDK devices" or something similar and put 3-pin devices (with substrate) in this group.

Radvall:
In addition to Resistor, Capacitor, and Diode, I've also moved the previously added BJT and MOS devices. I chose not to include all possible variations of elements in the 'PDK devices' group, as it was done for MOS transistors.

ra3xdh (approved):

ra3xdh:

Issue #193, opened 2022-12-06
Qucs-S not passing Resistor Tc1 or Tc2 to ngspice


tomhajjar:
Qucs-S doesn't seem to pass Resistor Tc1 or Tc2 to ngspice.

Another issue is choosing the variable name for "Temperature" when doing a Temperature Sweep. Tsweep isn't accepted and Temp doesn't do anything. 
 
![2022-12-05_190452](https://user-images.githubusercontent.com/102838604/205773620-dad749c1-ad88-4012-a462-a4413c0c5eee.jpg)

![2022-12-05_190847](https://user-images.githubusercontent.com/102838604/205773627-309e9dce-bf6e-4b67-b276-05bf473451d5.jpg)

https://sourceforge.net/p/qucs/discussion/311050/thread/82256003/

[Rdiv_tsweep.zip](https://github.com/ra3xdh/qucs_s/files/10159168/Rdiv_tsweep.zip)

ra3xdh:
Hello Tom, the Tc1 and Tc2 parameters of the blue resistors are ignored by Ngspice. Use red resistor instead. It allows to pass the full SPICE definition including model and temperature coefficients to the simulator. See the section 3.3.1 of the Ngspice manual for more details. The approach with `Tsweep` variable will not work with Qucs-S+Ngspice, because Ngspice cannot sweep variables. This will be improved after implementing of #189 in the upcoming v1.0.1 release. Use `TEMP` variable instead. It is a SPICE special global variable representing temperature.

ra3xdh:
Look at the attached screenshot for simulation example:
![image](https://user-images.githubusercontent.com/4920080/205915693-da8dd298-6393-43a0-a9ae-264e24d84ba6.png)

tomhajjar:
Should a comment be added to the device Description for devices that only pass parameters to Qucsator? 

![2023-01-16_093703](https://user-images.githubusercontent.com/102838604/212704112-d584e7d9-ffa6-48a4-a953-8fb6545cb9ed.jpg)

PR #434, opened 2024-01-05
Add resistor's parameters tc1 and tc2 to ngspice

Radvall:
I noticed that some people find it more convenient to change the parameters of passive components (resistor, capacitor, inductor) in the window 'Edit Component Properties', rather than using red components and specifying parameters in text.

I did some research on what parameters of passive blue components are supported to work with ngspice:

**capacitor**:

|Parameter|ngspice|
|---|---|
|capacitance in Farad (C) |✔|
|initial voltage for transient simulation (V)|✔|

**inductor**:

|Parameter|ngspice|
|---|---|
|inductance in Henry (L) |✔|
|initial current for transient simulation (I)|✔|

**resistor**:

|Parameter|ngspice|
|---|---|
|ohmic resistance in Ohms (R) |✔|
|simulation temperature in degree Celsius (Temp)|✖|
|first order temperature coefficient (Tc1)|✖|
|second order temperature coefficient (Tc2)|✖|
|temperature at which parameters were extracted (Tnom)|✖|

**Inductor with Q**

|Parameter|ngspice|
|---|---|
|Inductance (L) |✔|
|Quality factor (Q)|✔|
|Frequency at which Q is measured (f)|✔|
|Q frequency profile (Mode)|✔|
|simulation temperature in degree Celsius (Temp)|✖|

**Capacitor with Q**

|Parameter|ngspice|
|---|---|
|Capacitance (C) |✔|
|Quality factor (Q)|✔|
|Frequency at which Q is measured (f)|✔|
|Q frequency profile (Mode)|✔|
|simulation temperature in degree Celsius (Temp)|✖|

Thus since most of parameters for inductor and capacitor are supported, then they should also supported for the resistor, which is not true. 

And although this has already been discussed in the #193 I decided to add parsing temperature parameters for resistor. Also I added a comment 'Qucsator only' in the parameter Description for parameters that are ignored by Ngspice. 

Simulation results for blue and red resistors are shown bellow:

temp_res

### DIFF
--- a/qucs/components/capq.cpp
+++ b/qucs/components/capq.cpp
@@ -71,7 +71,7 @@ CapQ::CapQ()
 		QObject::tr("Q frequency profile")+
 		" [Linear, SquareRoot, Constant]"));
   Props.append(new Property("Temp", "26.85", false,
-                QObject::tr("simulation temperature in degree Celsius")));
+                QObject::tr("simulation temperature in degree Celsius (Qucsator only)")));
 }
 CapQ::~CapQ()
 {
--- a/qucs/components/indq.cpp
+++ b/qucs/components/indq.cpp
@@ -72,7 +72,7 @@ IndQ::IndQ()
 		QObject::tr("Q frequency profile")+
 		" [Linear, SquareRoot, Constant]"));
    Props.append(new Property("Temp", "26.85", false,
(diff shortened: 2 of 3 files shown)


ra3xdh:

Issue #124, opened 2022-07-29
Save .OP simulation result to dataset


tomhajjar:
Is there a reason Qucs-S won't do a simple DC operating point (.op)? Qucs does it and I assume ngspice will do it as well. 

If I add say a DC Parameter Sweep it does work.  

[Test_prj.zip](https://github.com/ra3xdh/qucs_s/files/9222783/Test_prj.zip)

ra3xdh:
I know about this problem. Qucs-S currently supports only F8 mode to show the .OP result. The single DC simulation device has no effect and the operation point is not saved to the dataset. See also #3

ra3xdh:
Combined with #3

ra3xdh:
Fixed via #513

PR #435, opened 2024-01-07
Add parser for printed scalars output

ivandi69:
In custom simulations one can print measurement or calculation results.
For example
```
ac lin 100 1meg 10meg
write custom.plot v(out)
meas ac fcenter max_at v(out)
let vmax=vecmax(v(out))
print fcenter vmax > custom.print
```
The custom.print file looks like this:
```
fcenter = 1.234e+01
vmax = 5.678e+00,4.321e-01

```
The parser adds printed results to the dataset as scalars.


ra3xdh:
Thanks for the contribution! This addition will help to implement the `.MEASURE` support and the `saveOp` option #124

PR #444, opened 2024-01-09
Refactor schematic scaling and resizing logic

wawuwo:
Hi @ra3xdh

Finally I can present it :)

As you've probably seen the schematic's zooming and scrolling logic is quite convoluted and scattered across a bunch of methods. This is an attempt to refactor it and make it more streamlined and simple. 

First two commits lay the foundation for zooming, scrolling, resizing, etc. by introducing a _conceptual model_ of rendering (something alike to MVC) and a _declarative approach_ to do it. Other commits just switch already known actions to using new rendering approach.

The _conceptual model_ is the following: every schematic element "exists" in some abstract cartesian coordinate system, a rectangular area of this system which encloses the elements is a _model_. A widget canvas on which schematic is drawn is the _view_. And the `renderModel` method is the _controller_ – size and position calculations are confined within this method.

The _declarative approach_ enables to describe a model bounds, a point within the model and where to draw it on a screen, pass this parameters to renderer and see the result.

These two ideas allow to get rid of cryptic code to compute new size and position.

While changes are mostly internal, users will also see some differences if this PR is merged:
1. "Zoom In" tool follows the click, like when zooming using Ctrl and mouse wheel
2. "Show all" positions the view so that schematic center is shown at viewport center
3. "Zoom to selection" positions the view so that the center of selected elements bounding rectangle is shown at viewport center
4. "Show 1:1": keeps the point in the viewport center stable while changing scale if the point is within a schematic bounding rectangle. Otherwise it centers the view on schematic's center.

### DIFF
--- a/qucs/mouseactions.cpp
+++ b/qucs/mouseactions.cpp
@@ -1946,31 +1946,11 @@ void MouseActions::MReleaseZoomIn(Schematic *Doc, QMouseEvent *Event)
 
     MAx1 = Event->pos().x();
     MAy1 = Event->pos().y();
-    float DX = float(MAx2);
-    float DY = float(MAy2);
 
-    float initialScale = Doc->Scale;
-    float scale = 1;
-    float xShift = 0;
-    float yShift = 0;
-    if ((Doc->Scale * DX) < 6.0) {
-        // a simple click zooms by constant factor
-        scale = Doc->zoom(1.5) / initialScale;
 
-        xShift = scale * Event->pos().x();
-        yShift = scale * Event->pos().y();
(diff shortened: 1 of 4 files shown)


wawuwo:
I've found a bug, please do not merge :(
I'll fix and force push an updated branch.

ra3xdh:
@wawuwo , Is this PR ready for review now?

wawuwo:
@ra3xdh yes

ra3xdh:
Thanks for the contribution! Merged. I hope this will help to improve the sources readability and maintenance.

Issue #437, opened 2024-01-07
Paintings, nets or wires with Dimension of "0"


tomhajjar:
As outlined in #426:

Attached is a schematic with a line and ellipse with a dimension of "0". If these end up "off screen" they both screw up "Zoom All" AND they cannot be selected for deletion. Only a text editor will find and delete them.

Shouldn't the ability to make any "Painting" like a line, ellipse, circle..etc with dimensions of "0" be removed.

![2024-01-07_115249](https://github.com/ra3xdh/qucs_s/assets/102838604/7f511e3b-0af7-4310-86fe-b5649465d13c)

[Test_prj.zip](https://github.com/ra3xdh/qucs_s/files/13854249/Test_prj.zip)

ra3xdh:
Yes, the best solution would be to disable zero dimensions everywhere.

wawuwo:
I think I'll make a patch today

tomhajjar:
"adds 1 to the element's bounding rect"

I assume this doesn't actually alter the dimension

ra3xdh:
The best solution would be to add the check in `MouseResizeMoving()` method in classes derived from `Painting`. This check should not allows to create zero size object and set the minimal size to 1x1.

wawuwo:
@tomhajjar I've made a patch and it seems to work nice, at least with a file you attached to this issue.
![zero-size-select](https://github.com/ra3xdh/qucs_s/assets/40355531/4838c396-0b5c-4064-a057-fa469b4ecbad)

tomhajjar:
No wire, painting or object of size "0" should be allowed

wawuwo:
@tomhajjar do you have any other "problematic" files to test the patch? Alternatively, you can build from the patched branch and test yourself :) If it's OK I'll open a PR

tomhajjar:
I can't compile Qucs so I wait for Vadim to create a Windows version...

wawuwo:
OK, I'm opening a PR then

PR #445, opened 2024-01-09
Fix zero-size elements selection

wawuwo:
Rewrite element selection logic in a way that allows to select "invisible" zero-size elements.

Fixes ra3xdh/qucs_s#437


ra3xdh:

PR #453, opened 2024-01-13
Add Chinese translation

hvwyl:
Chinese translation for Qucs-S


ra3xdh:
Thanks for the translations! Merging.

PR #455, opened 2024-01-13
Fix: select-all does not select wires on the boundary

hvwyl:
When the wire is on the boundary of content, use method "QRect::contains" will not select linear element on the boundary. Expand the area of "sizeOfAll" to fix it.

IMG_20240113_201812


ra3xdh:

PR #457, opened 2024-01-13
Uniformize 'capacitance with quality factor' name

ivandi69:
For IndQ we have **LQx** and **RLQx** in the generated netlist.
For CapQ we have **CCAPQ** and **RCAPQ**.
This patch changes CapQ netlist names to **CQx** and **RCQx**.


ra3xdh:

PR #473, opened 2024-01-20
Add a marker indicator at the end of the leader line.

iwbnwif:
The current marker implementation has a leader line from the values box to the point on the graph where the actual value is situated. This change provides an option to emphasize the marked point by drawing a box around it or adding a triangle with its lower point at the marked value.


ra3xdh:

PR #485, opened 2024-01-22
Graph limits by mouse

iwbnwif:
This is an initial PR for improvements to the diagrams to allow user interaction directly in the schematic. 
This first step makes it possible to 'zoom in' on a diagram by making a selection. In effect, this sets the limits that are already implemented in the diagram dialog.
Currently, only rectangular diagrams are supported.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,8 @@ qrc_*.cpp
 *.dat.spopus
 .idea/
 __pycache__
+.vscode/c_cpp_properties.json
+.vscode/launch.json
+.vscode/settings.json
+.vscode/tasks.json
+.gitignore
--- a/qucs/bitmaps/svg/viewwave.svg
+++ b/qucs/bitmaps/svg/viewwave.svg
@@ -1,0 +1,554 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="128"
+   height="128"
+   version="1.0"
+   id="svg171"
+   sodipodi:docname="viewwave.svg"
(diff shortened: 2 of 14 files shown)


zergud:
1. single click on the diagram in "set limits" mode gives "strange" results
2. if no graph falls within the rectangle - "strange" results

I suggest checking whether the graph falls within selection rectangle

zergud:
also, i get error "exp10() undeclared" on my PC
for testing replace to pow(10, x) 
suggest to use pow(10, x)!

iwbnwif:
> also, i get error "exp10() undeclared" on my PC for testing replace to pow(10, x) suggest to use pow(10, x)!

Sorry, yes this is a gcc extension - I will replace with pow(10, x) as you suggest.

iwbnwif:
> 1. single click on the diagram in "set limits" mode gives "strange" results

I have added a check for a minimum size for the selection. Previously, a single click generated a selection rectangle 1x1px so went to max zoom.

> 2. if no graph falls within the rectangle - "strange" results
> I suggest checking whether the graph falls within selection rectangle

Actually, there were already checks for a) starting in a rectangular diagram; and b) finishing in the same diagram as when the dragging started. It is just that the display wasn't refreshed on a failure! It does refresh now, so the strange results have gone, but there is no warning to the user. Hopefully, this is okay?

zergud:
@iwbnwif thank you for contribution

zergud:
I'll add your commit without PR

zergud:
found another bug
rigth axis not scaling ( variable zAxis  %) )
reopen issue, make new PR with right-clicking fix to

PR #486, opened 2024-01-23
Fix remaining dialogs with parent=0 or no parent

ivandi69:
In sway tiling window manager these dialogs appear as maximized top-level windows.
Setting the schematic pointer as parent pointer makes them appear as normal floating dialogs.


ra3xdh:

PR #488, opened 2024-01-23
Add "pan with mouse" feature

wawuwo:
Hello!

If merged this PR would add the ability to scroll schematic by moving mouse with middle button pressed down:
[panning.webm](https://github.com/ra3xdh/qucs_s/assets/40355531/788697ea-4e4e-4447-921c-2138509f8bf6)

It also introduces a change that I'm not quite sure about: I had to delete `setHidden(true)` and `setHidden(false)` before and after `resizeContents` in `renderModel` method because the `Q3ScrollView` contents widget didn't receive MouseMove events after hiding and then unhiding the widget. I haven't observed any differences after removing these lines though. May be they don't make any sense now (they were inherited from legacy implementation) and it's OK to remove them.

### DIFF
--- a/qucs/schematic.cpp
+++ b/qucs/schematic.cpp
@@ -530,6 +530,26 @@ void Schematic::PostPaintEvent(
 void Schematic::contentsMouseMoveEvent(QMouseEvent *Event)
 {
     emit signalCursorPosChanged(Event->pos().x(), Event->pos().y());
+
+    // Perform "pan with mouse"
+    if (Event->buttons() & Qt::MiddleButton) {
+        const QPoint currentCursorPosition = contentsToViewport(Event->pos());
+
+        if (const int dx = currentCursorPosition.x() - previousCursorPosition.x(); dx < 0) {
+            scrollRight(std::abs(dx));
+        } else if (dx > 0) {
+            scrollLeft(dx);
+        }
+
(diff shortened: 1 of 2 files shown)


zergud:
thank you for contribution!

ra3xdh:
@wawuwo , Thanks for the contribution! It would be good to place a "Pan" button on the toolbar near to "zoom" buttons. This will make the new feature visible for users.

zergud:
@ra3xdh pan by pressed down middle button does not require button in toolbar
it is "well known" behavior for EDA end CAD users

wawuwo:
@ra3xdh 
> It would be good to place a "Pan" button on the toolbar near to "zoom" buttons

Yeah, I see what you mean, but unfortunately it cannot be done just by adding a button :(

If I have correctly understood you, the "Pan" button you mean assumes that by clicking on it the user switches to a "panning mode", which has its own mouse event handlers; and when the user is done panning the previous "mode" has to be restored. I believe I know how to implement this _in general_, but I'm afraid diving into details may take some time. 

If this feature is really needed I think I will be able to work on it, but as @zergud wrote, "press mouse wheel to pan" is a common approach. Hence maybe there is no need for separate panning mode/button and users will discover the feature intuitively by themselves

ra3xdh:
> but unfortunately it cannot be done just by adding a button :(

If it requires a sufficient effort to implement, the task may be omitted. The current implementation will be sufficient.

PR #500, opened 2024-01-25
Fix assertion fails in 'renderModel'

### LINKED ISSUE #497: Assertion error when starting a new schematic
This issue relates to the development branch, e.g., statusbar-improvements, but is probably due to the MVC refactoring of the schematic layout.

Steps to reproduce:
1. Open Qucs-S
2. Without panning, scrolling, or zooming the blank main window, drag a lumped element onto the middle of the schematic
3. An assert occurs in Schematic::renderModel (line 2745 of the mentioned branch).

wawuwo:
This is an attempt to fix #497.

Using the build from this branch I tested the following cases on new empty schematic and also on schematic from a file.
1. Place an element in the middle of the view
2. The same as 1 for every corner
3. Zoom in and zoom out
4. Scroll and pan

I haven't observed any oddities or crashes.

### DIFF
--- a/qucs/schematic.cpp
+++ b/qucs/schematic.cpp
@@ -66,6 +66,19 @@ Q3PtrList<Node> SymbolNodes;
 Q3PtrList<Diagram> SymbolDiags;
 Q3PtrList<Component> SymbolComps;
 
+/**
+    If \c point does not lie within \c rect then returns a new
+    rectangle made by enlarging the source rectangle to include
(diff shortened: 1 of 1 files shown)


iwbnwif:
Thanks, this has solved the assert I was getting. I have used it for an hour or so and have not noticed any new issues.

ra3xdh:
Thanks for solution! Merging.

Issue #396, opened 2023-12-09
Get rid of digital simulation wrapper scripts


ra3xdh:
Digital simulation is launched using the wrapper script like `qucs_run_verilog`. It's need to switch to QProcess instead and provide a user setting to define `iverilog` executable location.

Issue #395, opened 2023-12-09
Replace FreeHDL by GHDL


ra3xdh:
Qucs-S uses FreeHDL for VHDL simulation. The FreeHDL is deprecated and doesn't work on modern platforms anymore. It's need to replace it with GHDL. See #393 

- [x] Test and merge Linux implementation #503 
- [ ] Provide BAT files for Windows. 
- [x] Revise VHDL models to ensure GHDL compatibility #512

zergud:
Qucs/qucs#830

ra3xdh:
It seems all VHDL models require a revision after switching to GHDL. I god a simulator freeze trying to simulate simple schematic containing one demultiplexer device. GHDL gives warning about infinite loop. The same schematic could be simulated with iVerilog without warning. Simple schematics containing logic gates work without problem using GHDL. FreeHDL is crashing on every simulation on my machine, so I cannot compare the result. 

Unfortunately, digital VHDL/Verilog design is not my field of expertise. The help in VHDL models revision is highly welcome.  

![image](https://github.com/ra3xdh/qucs_s/assets/4920080/f4372ba5-d270-4669-bae6-a271021175e3)
![image](https://github.com/ra3xdh/qucs_s/assets/4920080/dbe66fd8-dd38-4ff0-9c6f-0587c44b8a8c)

iwbnwif:
I don't pretend to know much about ghdl, but I think that each process must have some wait statement to give the simulator a chance to actually output something (unless it has a sensitivity list as is the case of the process which represents the demux block). 

In your example can easily work around this problem by appending "wait for 1 ns;" before the "end process;" in the Logic_0 process definition.

Having a quick look, it seems that none of the digital components that generate a pattern do this properly except for the 'digital source' which do to create the timing of their output toggle.

Simplistically, we could hard-code a "wait for 1 ns;" at the end of each relevant vhdlCode() function, e.g. in pad2bit.cpp, pad3bit.cpp, pad4bit.cpp, logic_0.cpp, and logic_1.cpp.

iwbnwif:
A quick test indicates the only problem remaining is with the R/S flip flop.
![image](https://github.com/ra3xdh/qucs_s/assets/11036083/c2be6879-2e06-4735-a8c4-711ee5f30df0)

iwbnwif:
R/S flip-flop is working fine - it just didn't like both inputs being driven from the same source (e.g., undefined behavior at 1ns transitions)!

[...]

PR #503, opened 2024-01-26
Use GHDL instead of FreeHDL for digital simulation [Linux Only]

iwbnwif:
The approach taken follows the advice given at Qucs/qucs#830. 
Includes syntax fixes to accommodate the slight differences between GHDL and FreeHDL.
It is not the full solution to #395 and does not address #396, but it allows basic digital simulation as previously.


ra3xdh:
Thanks for the contribution! Merged. The remaining thing is to update the Windows BAT files until #396 will be implemented.

ra3xdh:
> but updating the .bat script should be trivial thereafter

I will try to update the BAT myself in the next week.

Issue #227, opened 2023-02-20
Schematic glitches fedora 37 [Qt6]


aviadb:
In the schematic viewer,
moving parts or wires makes the wires to disappear until the component is released.

 
![215077901-7948cda4-bbdb-43e4-9e4a-ad628fac43ac](https://user-images.githubusercontent.com/13761044/220174543-76486a83-a275-4d44-93fd-142ce23cf4f9.gif)

Fedora 37, Qt6, KDE desktop with dark theme

ra3xdh:
It was already reported here https://github.com/ra3xdh/qucs_s/issues/190#issuecomment-1406371905 I remember about this issue, but unfortunately I cannot reproduce and debug it. I have tried Fedora installation into the VM and everything works as expected here. Physical installation of the Ubuntu22.04 with Qt6 also works as expected. I only may suspect this may be related to Wayland.

cznwhale:
The same behavior is on the raspberry pi version. (I use the September 22nd 2022 release, 32 bit version, Kernel: Linux 5.15.84-v7l+, on a 4B model)

ra3xdh:
The issue also may be fixed by https://github.com/ra3xdh/qucs_s/pull/214/commits/fe135fa8d45d649118ad6ce4934f781c6d8b9b55 Did you try compile the latest `current` branch?

cznwhale:
in my case (Raspberry Pi version) I used the precompiled binary (qucs-s_1.0.1-1_armhf.deb)

cznwhale:
I now realize that the same behavior is for Windows portable version (qucs_s_win64_1_0_1.zip)

ra3xdh:
I have just tested the Windows v1.0.1 package using the physical Windows installation (not VM) and also cannot reproduce the issue. Component dragging works as expected on my hardware. Therefore I cannot provide a fix, because I still cannot reproduce this bug.

aviadb:
Tried with the latest, problem still exists.
Any log dump that I can provide to help you in analyzing this behavior?

ra3xdh:
[...]

PR #509, opened 2024-01-30
Move by mouse - another implementation 

### LINKED ISSUE #505: Components are binding to wires during move
subj, make impossible to move component inside loops without sticking it to wires.

Probably links should be created after component is released above wire.

qucs-links

zergud:
fixes #505 
for simulations and equations show full bounding box instead of only header


ra3xdh:
Thanks for providing a fix for #505 . Is #227 remains not fixed now?

Issue #271, opened 2023-05-22
Run DC bias simulation if only DC simulation presents in the schematic


ra3xdh:
I still getting reports that simulator fails to start if the schematic contains only DC simulation. See #267 and #269. The DC simulation has no effect by itself for Qucs-S+Ngspice. This feature is documented in the new tutorial, but it seems the users still use old tutorial from Qucs. I propose to check the schematic and run DC bias simulation mode (F8) automatically if schematic contains only DC simulation. This may prevent such false positive error reports.

ra3xdh:
I have added this feature on the `current` branch. Now if schematic contains DC simulation only DC bias simulation mode (F8) will be initiated automatically. This should prevent future false positive bugreports. Closing this.

PR #513, opened 2024-01-31
Ngspice netlist rework, fixes #463

### LINKED ISSUE #463: Proper use of FOUR and FFT with AC and Trans
I cannot get AC, Trans and FFT data in the same analysis while AC, Trans and FOUR works. 

Compare C300_FFT to C300_FOUR plots:

1. For C300_FFT, ac.v(in), ac.v(out) and frequency are listed twice. "frequency" has 301 and 8193 data points. How is Qucs storing the AC versus FFT data? 

2. Why does FFT use the same start/stop as the AC analysis? FOUR doesn't have this restriction. I can set the start and stop.

3. The FFT plot is garbage unless I delete the AC analysis. 

C300_FFT

C300_FOUR

[Amp_prj.zip](https://github.com/ra3xdh/qucs_s/files/13944020/Amp_prj.zip)

ivandi69:
This rework fixes #463 using the recently added custom prefix feature.
Prefixing is activated only if conflicting simulations are found, so **the existing schematics are not affected**.

Every simulation has its own output file named spice4qucs.<simulation_name>.extension, ex: spice4qucs.ac1.plot.
If any conflicting simulations are found, ex: AC and FFT, prefixing is activated and when the dataset is built the <simulation_name> prefix applies to all simulations. If a conflicting simulation is added to a schematic the existing diagrams have to be updated with the new prefixed variables. 

Other changes:
- Nutmeg equation now uses a complete or partial simulation name. For example if 'SP' is specified as simulation name the equation is included in all SP simulations and if SP1 is specified it is included in SP1 only. if ALL is specified it is included in all simulations.
- If a DC simulation is placed in the schematic the operation point is saved in the dataset.

Here are some tests:
[463-tests.zip](https://github.com/ra3xdh/qucs_s/files/14106663/463-tests.zip)

### DIFF
--- a/qucs/components/dc_sim.cpp
+++ b/qucs/components/dc_sim.cpp
@@ -88,7 +88,13 @@ Element* DC_Sim::info(QString& Name, char* &BitmapFile, bool getNewOne)
   return 0;
 }
 
-QString DC_Sim::spice_netlist(bool)
+QString DC_Sim::spice_netlist(bool isXyce)
 {
-    return QString("");
+    QString s;
+    if ( !isXyce ) {
+        s += "op\n";
+        s += QString("print all > spice4qucs.%1.ngspice.dc.print\n").arg(Name.toLower());
+    }
+
+    return s;
 }
--- a/qucs/components/equation.cpp
+++ b/qucs/components/equation.cpp
@@ -190,7 +190,8 @@ QString Equation::getEquations(QString sim, QStringList &dep_vars)
                 int idx = ng_vars.indexOf(Props.at(i)->Name);
                 if (idx>=0) used_sim = ngsims.at(idx);
             }
-            if ((sim == used_sim)||(used_sim=="all")) {
+            if ( used_sim.toLower() == "tran" ) used_sim = "tr";
+            if ( (sim.startsWith(used_sim.toLower())) || (used_sim=="all") ) {
                 eqn = tokens.join("");
                 s += QString("let %1=%2\n").arg(Props.at(i)->Name).arg(eqn);
                 dep_vars.append(Props.at(i)->Name);
--- a/qucs/components/param_sweep.cpp
+++ b/qucs/components/param_sweep.cpp
@@ -206,7 +206,7 @@ QString Param_Sweep::getNgspiceAfterSim(QString sim, int lvl)
 
     s = "set appendwrite\n";
 
-    if (lvl==0) s += QString("echo \"$&number_%1  $%2_act\">> spice4qucs.%3.cir.res\n").arg(par).arg(par).arg(sim);
+    if (lvl==0) s += QString("echo \"$&number_%1  $%2_act\" >> spice4qucs.%3.cir.res\n").arg(par).arg(par).arg(sim);
     else s += QString("echo \"$&number_%1\" $%1_act >> spice4qucs.%2.cir.res%3\n").arg(par).arg(sim).arg(lvl);
     s += QString("let number_%1 = number_%1 + 1\n").arg(par);
 
@@ -249,7 +249,7 @@ QString Param_Sweep::spice_netlist(bool isXyce)
         step = (stop-start)/points;
     }
 
-    if (Props.at(0)->Value.startsWith("DC")) {
+    if (Props.at(0)->Value.toLower().startsWith("dc")) {
         QString src = getProperty("Param")->Value;
         s = QString("DC %1 %2 %3 %4\n").arg(src).arg(start).arg(stop).arg(step);
         if (isXyce) s.prepend('.');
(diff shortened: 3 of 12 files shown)


ra3xdh:
Thanks for the contribution! I will test it in the next few days and post my feedback. I have seen also fix for #124 in this issue (export DC OP to the dataset).

qucs/components/dc_sim.cpp
```diff
@@ -88,7 +88,13 @@ Element* DC_Sim::info(QString& Name, char* &BitmapFile, bool getNewOne)
   return 0;
 }
 
-QString DC_Sim::spice_netlist(bool)
+QString DC_Sim::spice_netlist(bool isXyce)
 {
-    return QString("");
+    QString s;
+    if ( !isXyce ) {
```
ra3xdh, line 94: Is this intended to be fix for #124?

ivandi69:
Yes, but for now it works if DC is **not** the only simulation. I have to figure out how to handle F8.

ra3xdh:
I will remove checker error when the schematic contains DC simulation only. The F8 mode (DC bias) may be left unchanged. Qucsator doesn't export data to dataset using F8 mode too.

ra3xdh:
I have found an issue with Nutmeg equations. Nutmeg equations are not attached to SPICE simulation now. This may lead to computation error and simulation failure. For example look at the attached schematic. The equation for `K` causes division by zero error when executing TRAN simulation. The intention of the `simulation` parameter of the Nutmeg equation was to enable short variable notation like `v(node)` and exclude errors like the mentioned above. The equation for `K` is valid for AC but not valid for TRAN. Therefore I decided to attach every Nutmeg equation to specific simulation. 

image
image

ra3xdh:
I understood how new Nutmeg equation works. It is required to enter the simulation name in the `Simulation` field manually. This may be not user-friendly. The best solution would be to add a custom dialog with drop-down list for this device. .Unfortunately, there is no way to fill the drop-down list dynamically for such fields. But this will require a significant time effort. At least a new way of attaching Nutmeg equation to the simulation should be documented.

ivandi69:
> Nutmeg equation now uses a complete or partial simulation name. For example if 'SP' is specified as simulation name the equation is included in all SP simulations and if SP1 is specified it is included in SP1 only. if ALL is specified it is included in all simulations.

A drop-down list wont allow entering partial simulation name that applies to all simulation of the same type, SP for example. One can have an equation 'db_s21=db(s_2_1)' that applies to all SP simulation and another equation that applies to a particular simulation SP2. A combo-box maybe, I have to look how SWeep does it.

ivandi69:
Just found an ussie, '.INCLUDE ngspice_mathfunc.inc' is the first line in the netlist. Above '* Qucs 24.1.0' line. I removed the '* Qucs 24.1.0' duplication and it was OK, but now I see that it was because there was no ngspice_mathfunc.inc file in the test build. Will fix it as soon as I have some spare time.

ivandi69:
> I will remove checker error when the schematic contains DC simulation only.

https://github.com/ra3xdh/qucs_s/pull/513/commits/0625e1eaef5842f444bc3a2b95a7a15a3f5b3948 does it.

ra3xdh:
> A drop-down list wont allow entering partial simulation name that applies to all simulation of the same type,

I have inspected the `componentdialog.cpp` code. It should be possible to update the existing dialog without implementing a special dialog for Nutmeg equation. I propose the following logic.

* Introduce a keyword like `{SIMULATION}` in the property description. If Qucs-S finds this keyword, the drop-down list will be shown instead of line edit.
* The dialog looks into the parent schematic of the component and finds all simulation devices. The devices name and `ALL` keyword are added to the drop-down list. 
* If more than one simulation of the same type was found (like AC1 and AC2), the keyword like AC (simulation name without numbers) should be added to the list.

ra3xdh:
> https://github.com/ra3xdh/qucs_s/commit/0625e1eaef5842f444bc3a2b95a7a15a3f5b3948 does it.

Sorry, I forgot. It's need to revert this commit too https://github.com/ra3xdh/qucs_s/commit/0625e1eaef5842f444bc3a2b95a7a15a3f5b3948 does it. See https://github.com/ra3xdh/qucs_s/issues/271 Now if schematic contains only DC it is always switched to DC bias mode.

ra3xdh:
> https://github.com/ra3xdh/qucs_s/commit/0625e1eaef5842f444bc3a2b95a7a15a3f5b3948 does it.

Yes, the DC simulation works as expected now. It makes sense to remove the `AbstractSpiceKernel::checkDCSimulation`

ra3xdh:
I have added TODO list #518

ra3xdh (approved):

ra3xdh:
I have tested this PR with my simulations and found no major issues. I am merging this. The improvement of the component properties dialog could be implemented later. @ivandi69 Please stay connected if some issues related to this PR will raise in the next weeks. And thanks for good job.

tomhajjar:
I updated the ngspice 100 watt amp example to work with the latest nightly release.

AC, Tran, Four and FFT in a single analysis. Will soon add thermal like original 100W.sp example.



[Amp_100W_ngspice_prj.zip](https://github.com/ra3xdh/qucs_s/files/14213783/Amp_100W_ngspice_prj.zip)

PR #519, opened 2024-02-01
Implement simulation selection list for Nutmeg equaiton

ra3xdh:
As proposed in #513, I have added a simulation selection list for Nutmeg equation. No keyword were required. @ivandi69 ?

image


ra3xdh:
Merging.

PR #521, opened 2024-02-01
Refactor schematic's grid-drawing

wawuwo:
Hi! Legacy implementation looked a bit cryptic with all those bit-shifts and one-letter variable names, so I decided to refine it a bit.  Changes are mostly about code readability and about how the method is called. An overview:

- rename method. "Draw" is better name than "paint" because "painting" is about "covering a surface with a color". We paint a wall, a house, a garage door, but we draw a circle, a line a point
- reduce number of arguments. There is no need to pass some of the arguments as their values are accessible from within the method
- pass ViewPainter by const reference instead of a pointer
- rewrite method body to make it less cryptic

I've tested it on my machine, IMO the grid looks the same :)

### DIFF
--- a/qucs/schematic.cpp
+++ b/qucs/schematic.cpp
@@ -433,7 +433,7 @@ void Schematic::drawContents(QPainter *p, int, int, int, int)
 
     Painter.init(p, Scale, -ViewX1, -ViewY1, contentsX(), contentsY());
 
-    paintGrid(&Painter, contentsX(), contentsY(), visibleWidth(), visibleHeight());
+    drawGrid(Painter);
 
     if (!symbolMode)
         paintFrame(&Painter);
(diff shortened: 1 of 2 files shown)


iwbnwif:
I would like to replace the hard coded pen width with a variable. The grid is barely visible on HiDPI (e.g. 4k) displays. I think Qt allows for fractional pen widths now, so it should be possible to have a good guess at a suitable width from the DPI value. I understand if it's beyond the scope of this PR though!

ra3xdh:
Thanks for the contribution! Merged. @wawuwo Are you planning to make more refactoring PRs in the near future? I am going to prepare a release package for v `24.1.0` after February, 15. And we need a bit beforeahead planning.

wawuwo:
@ra3xdh maybe I'll try to refactor the `paintFrame` method as its looks quite bad. I was also thinking about providing a decent documentation for the rendering facilities ("MVC-like") I introduced earlier, and adding the Doxygen support to cmake config.

February, 15 is the beginning of "freeze period", is it correct?

@iwbnwif 
>to replace the hard coded pen width

I'll look into it, may be there is a way to calculate the appropriate pen width based on screen resolution.
Also, as you have a very wide display, could you please check something for me? I think I have a mistake in grid-drawing logic which makes nodes in bottom-right corner to be drawn "out of sync". Try to place an element somewhere around bottom-right corner. Do element's ports "snap to" grid-nodes? Or they miss a tiny bit?

iwbnwif:
> Also, as you have a very wide display, could you please check something for me? I think I have a mistake in grid-drawing logic which makes nodes in bottom-right corner to be drawn "out of sync". Try to place an element somewhere around bottom-right corner. Do element's ports "snap to" grid-nodes? Or they miss a tiny bit?

They miss a tiny bit. This is at about 6000 physical pixels from the top left corner.
image

ra3xdh:
@wawuwo 

> February, 15 is the beginning of "freeze period", is it correct?

Yes, it's better to plan massive refactoring PRs like #444 after the release will be done.

wawuwo:
@iwbnwif oh, thank you! Could you please also check whether the patch from https://github.com/wawuwo/qucs_s/tree/fix-grid-slip helps?

PR #524, opened 2024-02-02
Fix poor grid appearence on HiDPI displays

wawuwo:
This is an attempt to fix the issue @iwbnwif told about in #521.
>The grid is barely visible on HiDPI (e.g. 4k) displays.

Previously pen width was set to 0 which effectively made it a "cosmetic pen" i.e. 1-pixel-wide.  In this patch pen-width is calculated based on output of `physicalDpiX()`.

### DIFF
--- a/qucs/schematic.cpp
+++ b/qucs/schematic.cpp
@@ -1059,10 +1059,14 @@ void Schematic::drawGrid(const ViewPainter& p)
     if (!GridOn)
         return;
 
+    // A grid drawn with pen of 1.0 width reportedly looks good both
+    // on standard and HiDPI displays.
+    // See here for details https://github.com/ra3xdh/qucs_s/pull/524
+    p.Painter->setPen(QPen{ Qt::black, 1.0 });
+
     {
         // Draw small cross at origin of coordinates
         const QPoint origin = contentsToViewport(modelToView(QPoint{0, 0}));
-        p.Painter->setPen(QPen(Qt::black, 0));
         p.Painter->drawLine(origin.x() - 3, origin.y(), origin.x() + 4, origin.y());  // horizontal stick
         p.Painter->drawLine(origin.x(), origin.y() - 3, origin.x(), origin.y() + 4);  // vertical stick
     }


iwbnwif:
Thank you, I was just about to work on something similar!

Actually, I think it is enough just to use a pen width of 1.0 everywhere instead of 0.  

I am curious if (for example) if changing line 77 of graph.cpp as follows still looks okay on standard screens?

` 
// **** not selected ****

p->Painter->setPen(QPen(QColor(Color), Thick * p->PrintScale + 1.0, Qt::SolidLine));
`

wawuwo:
@iwbnwif 
> I think it is enough just to use a pen width of 1.0 everywhere instead of 0

Have you tried it, does it look good on your screen? I mean QPen documentation [says](https://doc.qt.io/qt-5/qpen.html#setWidth) zero width pen is the same as 1-pixel width and you've mentioned that it is barely visible on your device… I tried 1-pixel width pen and on my display (1368x768 14'') grid does look the same as with zero-width pen :)

wawuwo:
Maybe `ViewPainter` clients shouldn't set pen width or color to concrete values. Instead they should set "thin" or "very thin", "main color" or "accent color" and so on and concrete values should be calculated and confined in `ViewPainter`. This would hide calculations of pen width and would allow to change color sets (dark theme!) without changing the client code. Decoupling is usually a good thing.  (Just a thought)

iwbnwif:
> Have you tried it, does it look good on your screen? I mean QPen documentation [says](https://doc.qt.io/qt-5/qpen.html#setWidth) zero width pen is the same as 1-pixel width and you've mentioned that it is barely visible on your device… I tried 1-pixel width pen and on my display (1368x768 14'') grid does look the same as with zero-width pen :)

Yes, 1.0 width pen looks much better, both on the grid and the diagrams / graphs. 

I read the documentation to mean 0 width means one _physical_ pixel which are tiny on a 4k display, whereas 1.0 is scaled appropriately.

Empirically, 1.0 looks good on my 4k, whereas 0 is not for the grid, and too thin for graph (variable) lines. 0 is just about okay for the diagram grid / scale.

Unfortunately, I don't have access to that computer at the moment so can't do any tests until tomorrow evening.

iwbnwif:
> Maybe `ViewPainter` clients shouldn't set pen width or color to concrete values. Instead they should set "thin" or "very thin", "main color" or "accent color" and so on and concrete values should be calculated and confined in `ViewPainter`. This would hide calculations of pen width and would allow to change color sets (dark theme!) without changing the client code. Decoupling is usually a good thing.  (Just a thought)

Yes, this is also a good idea. I also experimented in making it a user setting, especially the default graph line widths as I actually prefer 1.5 or 2.0 😉 for those. But I am learning at the moment and probably will create far fewer graphs in practice!

ra3xdh:
No problems found using HiDPI screens. Merging.

Issue #366, opened 2023-11-15
Bug in parsing Verilog-A code


dwarning:
In the process of symbol generation code lines in the .va model like

`// Process related parameters
    parameter real dl = -0.01E-6; // `P(info="Channel width correction" units="m");
    parameter real dw = -0.01E-6; // `P(info="Channel length correction" units="m");
`
or
`        /* vto, kp and gamma with variation for MC analysis if required.
         * The default value for model parameters avto, akp and agamma
`
follow in wrong parameters in the instance symbol like 
`process=related`
`default=value`
or similar.
Correct behaviour would be to ignore any commented line.

**TODO** list

- [ ] Exclude commented lines from parser
- [ ] Use preprocessor for Verilog-A files

ra3xdh:
I need the full `*.va` source file to debug this issue. It may be more complex than it  seems.

dwarning:
I attached the project in my #365 including the ekv26.va.

ra3xdh:
Yes, I confirm this. Qucs-S cannot recognize parameters in this file.

dwarning:
I looked into certain files and saw blocks for skipping comment blocks from va files, e.g. vafile.cpp. This seems not to work. Which file is used for reading the va file parameter names and values to include in the json files?
BTW - Because latest cmc models using macros for instance and model parameters I am looking for a general solution in transform these model through a preprocessor into an expanded format. Maybe openvaf --print-expansion would be a candidate, but there are missing LF in the output at the moment.
Your opinion?

dwarning:
May be it helps:
With the beta version of VAMPyRE (rel. in December) by Geoffrey Coram from CMC is it possible to preprocess any Verilog-A file. As a result you have a file with pure va text belong the VA standard 2.4 and striped any comment lines.
The parameter section will then have entries like, here summarized as examples:
`
parameter vto = 0.5; <- parameter default real type

parameter real vto = 0.5;

parameter real vto = 0.5 from( -inf : inf );

(*desc= "Vth"*) parameter real vto = 0.5 from( -inf : inf );

(*units= "V" , desc= "Vth"*) parameter real vto = 0.5 from( -inf : inf );

(*units= "C" , type="instance" desc= "device deviation temp"*) parameter real dtemp = 0.0 from( -inf : inf );
`

Then I used a slight modified function _Schematic::savePropsJSON_ 
`    QString module;
    QStringList prop_name;
    QStringList prop_val;
//    QStringList prop_desc;
[...]

Issue #411, opened 2023-12-26
Add entry field for external model for Verilog-A


dwarning:
Vadim,
I have a big wish.
The life would be much easier if there would an entry field for model cards which can provided by include a file. I made a sketch:
![Bildschirmfoto vom 2023-12-26 19-18-56_](https://github.com/ra3xdh/qucs_s/assets/18441932/14204681-d459-4437-b7a9-01c67771e2d2)
At the moment I have to change the parameters painful by editing the parameter entries. If I change the model anything is lost and we have the fallback to the Verilog-A default values.
My proposal is to add a checkbox and an entry field for a model name. If both exist the user have to provide the model card in a file which has to be included. In this case the modelcard from symbol have to omitted in the netlist.

dwarning:
No response so far. For me a show stopper because I see no chance to include model cards for VA models.
I will close.

ra3xdh:
> No response so far. 

The response may require some time, because I am developing Qucs-S in my spare time. Usually I am responding within 1 week, but some issues may require longer time. The project has no full-time developers. Please wait. 

> if there would an entry field for model cards which can provided by include a file

The implementation of your proposal is not so easy as it seems. The current design of verilog-a device is strongly dependent on JSON files. The symbol is separate JSON object. I would not touch this software solution, but rewrite it from the scratch. The adding of the modelcard reference may be not possible using the existing implementation. I am planning to implement Verilog-A file device that will act like "SPICE file" device and will not require the usage of JSON.  But the implementation of this feature has a low priority for me because of low significance of compact modelling for 99% of average Qucs-S users.  So, don't expect the new Verilog-A device will be added in this year.

But there exist a workaround for your request using the existing  "SPICE generic device component". You may aslos attach one model to many devices (for example N or P type) using this workaround. Perform the following steps:

* Compile the `*.va` file into the OSDI module. No matter if form Qucs-S of from CLI.
* Put "SPICE generic (Z)" device on schematic. Define the letter "N" and pin number. Fill the model reference field.
* Put ".MODEL" modelcard device on schematic. Fill the modelcard in it.   
* Put "SPICEINIT" device on schematic. Fill the "osdi <path_to_osdi>" command here. 
* Add simulations and other devices. Pack the "SPICE generic device" into subcircuit if necessary. Define symbol and forward subcircuit parameters into the modelcard. 

The attached screenshots illustrate this workaround for tunnel diode model.

[...]

PR #527, opened 2024-02-04
more precise parameter parsing for Verilog-A files

dwarning:
Related to #366:
This code change helps the parameter parsing of Verilog-A model files.
There are still other flaws, e.g. that any "parameter" string in comments follows in wrong symbol files.


ra3xdh:

ra3xdh:
The workflow proposed in #411 still be preferren for Nsgpice+OpenVAF. It doesn't require parsing of VA-file.

dwarning:
Thanks for merging!
Parsing of VA-file is needed for symbol generation: *_props.json and *_symbol.json

ra3xdh:
The approach proposed in #411 doesn't require JSON files. But this will not work for Qucsator+ADMS.

Issue #368, opened 2023-11-16
Implement shortcut manager


ra3xdh:
The feature requested by @medwatt There exists an unfinished shortcut manager PR for Qt4 version of Qucs https://github.com/Qucs/qucs/pull/558 It is need to evaluate its status and backport the code into Qucs-S or implement a new shortcut manger. I don't find this feature useful, but user demand on this exists. This may be added as a low priority task.

ra3xdh:
This task remains open more than one year. @zergud Do you have any progress on this or the task may be cancelled? It is not critical for the application operation and doesn't affect the simulation and I am not planning to implement this feature myself.

PR #529, opened 2024-02-04
Persist geometry of ComponentDialog

iwbnwif:
The ComponentDialog default size is generally too small to see all the rows and text.

This change 'remembers' the last size that the dialog was set to when the user pressed 'OK'.

It might be better to implement the settings write in the close event.

### DIFF
--- a/qucs/components/componentdialog.cpp
+++ b/qucs/components/componentdialog.cpp
@@ -41,7 +41,9 @@
 ComponentDialog::ComponentDialog(Component *c, Schematic *d)
 			: QDialog(d)
 {
-  resize(450, 250);
+  QSettings settings("qucs","qucs_s");
+  restoreGeometry(settings.value("ComponentDialog/geometry").toByteArray());
+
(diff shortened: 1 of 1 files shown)


iwbnwif:
I am thinking that it might be tidier to put the storing into onShow and onClose event handlers of a QucsDialog base class. Then any dialog that is resizeable can automatically inherit this functionality.

For 24.1.0 I would only propose to implement ComponentDialog from the new base class so we can be sure there are no issues.

zergud:
it would be nice to implement settingsManager or at least use static  QSettings settings("qucs","qucs_s") instead of hardcoded "organization" and "application" in many files
settigsManager is planned for #368 to store shortcuts for all actions

zergud:
all fuctionality + 
- load defaults
- export/import setting
at least for shorcuts


I think there is no need to rush and implement this in next release

@ra3xdh  this PR can be merged now :)

ra3xdh:

Issue #496, opened 2024-01-24
Text document (TextDoc) font is too small on HiDPI displays


iwbnwif:
TextDoc is the built in Qucs-S editor and it has a fixed font size that is set as the schematic font size - 1.

A reasonable setting for the schematic font size (that is in proportion to the symbols) is 10pt, but this creates a "Courier New" 9pt font for the text in TextDoc.

Is it worthwhile to add another option in Application Settings for the TextDoc font size? 

It may also be nice to allow mouse wheel zooming in TextDoc via QPlainTextEdit::zoomIn and QPlainTextEdit::zoomOut as a workaround for any edge cases or accessibility.

ra3xdh:
I have found one issue with this PR. If the text editor is already opened, the font size is not updated if the settings dialog is executed and `OK` pressed. It's need to close editor tab and open it again to update the font size. Is this expected behavior?

iwbnwif:
Yes, actually I followed the other font settings in the Applications Settings dialog and wrote "... (set after reload)" next to the setting.

I think this is only a minor inconvenience because I wouldn't expect users to change the default setting that often, but we could search through the open tabs to see if there is a text document and refresh it?

ra3xdh:
> I think this is only a minor inconvenience 

Yes, the things may be left as is.

PR #534, opened 2024-02-07
Implement mouse wheel zoom of text documents

iwbnwif:
This is a simple fix for #496 to allow mouse wheel zooming of text documents. It does not persist the zoom level between runs and any new documents are opened at the default zoom.

This PR can either be merged now, wait for 24.2.0, or wait for the settings refactoring #533 to implement persistence.

### DIFF
--- a/qucs/qucs.cpp
+++ b/qucs/qucs.cpp
@@ -1983,6 +1983,7 @@ void QucsApp::slotChangeView()
     Doc = (QucsDoc*)d;
     // update menu entries, etc. if necessary
     magAll->setDisabled(true);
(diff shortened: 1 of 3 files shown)


ra3xdh:
This could be added in the upcoming release. I have assigned `24.1.0` milestone for this PR.

ra3xdh:
I have tested the proposed PR and found no major issue. I would propose to add something like "Reset default zoom" in the text editor context menu.

ra3xdh:
Yes, I have found that "zoom 1:1" does zoom reset. No further changes needed. I am merging this.

PR #536, opened 2024-02-09
Add support for logarithmic sweep to DISTO simulation

ivandi69:
The code is borrowed from NOISE and the example is updated to show the sources and equations needed.


ra3xdh:

PR #539, opened 2024-02-09
Add clangd LSP support

wawuwo:
Hi! I think it would be nice to have clangd support "out of the box". I understand that maybe I'm the only one who thinks like that (ha-ha), so I will totally understand if PR won't be accepted :)

Changes:
- Alter cmake config so it generates `compile_commands.json`
- Add clangd config
- Add to README a section describing how to set up clangd LSP


ra3xdh:

PR #541, opened 2024-02-09
Add default TextDoc font to settings dialog

### LINKED ISSUE #496: Text document (TextDoc) font is too small on HiDPI displays

iwbnwif:
This change adds an extra entry to the settings dialog to allow the user to choose a default font for any documents opened in the text editor. The initial default value follows the system setting for a monospaced font.

I wasn't going to do this until after the settings refactoring in #533, but on the other hand it would be nice to close #496 before releasing 24.1.0.

### DIFF
--- a/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/dialogs/qucssettingsdialog.cpp
@@ -84,18 +84,23 @@ QucsSettingsDialog::QucsSettingsDialog(QucsApp *parent)
     connect(AppFontButton, SIGNAL(clicked()), SLOT(slotAppFontDialog()));
     appSettingsGrid->addWidget(AppFontButton,1,1);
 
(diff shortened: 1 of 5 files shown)


ra3xdh:
I have assigned `24.1.0` milestone for this issue. I will merge this after explanations on my comment above.

ra3xdh:
Merged.

Issue #469, opened 2024-01-19
2.1.0 GUI is overscaled on 4k displays (Win 11)


iwbnwif:
All elements of the Qucs-S GUI are too big when using Windows 11 display scaling.

This is a new issue in 2.1.0 and didn't occur in 2.0.0.

It looks as though the scaling factor is being applied twice. I have the scale set to 150% and everything is approximately 1.5 times bigger than it should be (i.e., compared to 2.0.0 or other Qucs's.

It is possible to work around the problem by creating the following VBS script:

```
Dim WshShell, WshEnv
Set WshShell = CreateObject("WScript.Shell")
Set WshEnv = WshShell.Environment("PROCESS")
WshEnv.Item("QT_SCALE_FACTOR") = "0.67"
WshShell.Run "qucs-s.exe
```

The GUI size is fine in Ubuntu on the same hardware and with the same desktop scaling.

ra3xdh:
Thanks for reporting and providing a workaround. I have never observed scaling issues with fullHD hardware using Linux or Windows platform. Unfortunately I have no access to 4k hardware to test what happens. Probably some scaling options should be applied as the `QApplication` attribute. The best way if someone will provide a patch fixing this issue.

zergud:
this is quite a complex problem and relates to  non-integer scale factors (such as 150%)
we have different behavior for qt5 and qt6

if set Qt::HighDpiScaleFactorRoundingPolicy::Floor in version 2.1 we obtain correct GUI scaling (menu toolbar)
but wrong font scaling in schematic (problem exists in earlier versions as well)

need futher investigation!

![image](https://github.com/ra3xdh/qucs_s/assets/869819/7c212a92-dda1-4671-b2ab-b4a77bb7f6bd)

iwbnwif:
I notice that the Windows binary ('portable package') for 2.1.0 is built using Qt5. I am not able to build on Windows at the moment, but would be happy to test a Qt6 version if there is a binary available.

wawuwo:
This ticket has been silent lately… Is GUI still overscaled on 4k in 24.1.0?

iwbnwif:
Yes, the binary available from the main site still has the overscaling problem.

zergud:
You need to add option - line 832
and test on 4k display

![IMG_20240229_000829_035](https://github.com/ra3xdh/qucs_s/assets/869819/6e99c03e-58c8-4fda-9092-ef991d29973a)

[...]

PR #542, opened 2024-02-09
Fix component font size when Qt::HighDpiScaleFactorRoundingPolicy is set

wawuwo:
@zergud, hi!

This is an attempt to fix the bug you described [there](https://github.com/ra3xdh/qucs_s/issues/469#issuecomment-1903477584). Hopefully you still have access to HiDPI hardware you were using when you discovered the bug. Could you please check whether this patch helps or not?

I open PR as a draft in case the patch doesn't fix the issue.

### DIFF
--- a/qucs/viewpainter.cpp
+++ b/qucs/viewpainter.cpp
@@ -53,7 +53,11 @@ void ViewPainter::init(QPainter *p, float Scale_, int DX_, int DY_,
 #ifdef __MINGW32__
   FontScale = Scale;
 #endif
-  f.setPointSizeF( FontScale * float(f.pointSize()) );
+  {
+    QFontInfo fontInfo{ f };
+    double fontSize{ static_cast<double>(fontInfo.pixelSize()) * FontScale};
+    f.setPixelSize( static_cast<int>(fontSize) );
+  }
   p->setFont(f);
   LineSpacing = p->fontMetrics().lineSpacing();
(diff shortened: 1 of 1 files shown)


zergud:
Hi!
I can check this patch in 2 days
if anyone can do it sooner, it would be nice )

zergud:
this patch helps!
blurry font in the menu is another problem )

p.s. why are there parentheses on line 56 and 60 (maybe I don't know something in C++)?

image

wawuwo:
@zergud 
>this patch helps!

This is very good news! I'll mark PR as read for review then.

>why are there parentheses on line 56 and 60 (maybe I don't know something in C++)?

I put those three statements in a dedicated block because they're closely relatated and to restrict the scope of `fontInfo` and `fontSize` because they aren't needed anywhere else. My intention was to enhance readability.

There is also a nice side-effect that block-variables are [destroyed ](https://en.cppreference.com/w/cpp/language/statements#Compound_statements) after leaving the block. So once `fontInfo` and `fontSize` have been used to set up font size, they aren't needed anymore and could be safely destroyed.

I'm not much of an C++ expert and I'm not 100% sure that it's a good thing do so, but at least it looks reasonable to me :)

zergud:
I assumed something like this
but method ViewPainter::init is not  "very complicated", and I think this is premature optimization and affects readability
In any case, until the coding standards are approved, this is just me grumbling ;)

PR #543, opened 2024-02-09
Update qucs_run_hdl.bat to use GHDL on Windows

iwbnwif:
*** UNTESTED! ***

This should complete #395 but I am not able to build and test Qucs-S on Windows at the moment. Please can someone test this before proceeding any further.

### DIFF
--- a/qucs/qucs_run_hdl.bat
+++ b/qucs/qucs_run_hdl.bat
@@ -25,13 +25,19 @@ if "X%6"=="X" goto usage
 
 set NAME=%1
 set NAMEOUT=%2
-set TIME=%~3
-set DIR=%4
+set SIMTIME=%~3
+set SIMDIR=%4
 set BINDIR=%5
 
-if not exist "%DIR%" goto nodir
+REM echo %0 %1 %2 %3 %4 %5
 
-cd /d "%DIR%"
+REM Remove any spaces in the simulation time command.
+set SIMTIME=%SIMTIME: =%
+REM echo %SIMTIME%
+
+if not exist "%SIMDIR%" goto nodir
+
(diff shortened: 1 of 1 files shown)


ra3xdh:
The BAT doesn't work for me. The error message is shown on the screenshot. I suspect the problem with `%TIME%` variable. Maybe my GHDL windows package is wrong, because the same circuit works using Linux package. You may use this demo package to test the GHDL on Windows https://drive.google.com/file/d/1-anZcfQK9heIn0d9nmqSGqHlPXDfOjUh/view?usp=sharing Everything should work using 64-bit wine. 
image

ra3xdh:
The simulation operates after setting `time=12n` without unit in the simulation properties. But GHDL shows an another error coming from the VHDL file itself. It's strange that the same file works on Linux without problem. 

image

iwbnwif:
Please can I check if the Windows build you sent has the fix to [qucs/dialogs/simmessage.cpp](https://github.com/ra3xdh/qucs_s/pull/503/files#diff-9d0583b83d6c7cfb6642fe1e86cce61b360f05846658a9ca3c79cb0fa8d6cb29) in faaaff032a82e2caa5ea1f1ddd378a79e202b30b? There should not be a space between '40' and 'ns'.

I have a workaround, but it would be better to test it against the final code.

But there are also some problems with the batch file. It seems that %DIR% and %TIME% are reserved - at least on my Wine installation. If I change them to %SIMDIR% and %SIMTIME% it seems to work.

ra3xdh (approved):

ra3xdh:
Everything works as expected now. Merging.
image

Issue #94, opened 2022-07-01
Add SPICE models for transmission lines devices


ra3xdh:
The S-parameter simulation support for Ngspice and Xyce backends has been added in the 0.0.24 release. But the SPICE entries for transmission line devices including microstrip still be missing. These devices are unique for Qucsator and have no direct SPICE equivalent. It's need to provide a way to simulate old Qucs RF-circuits with new Ngspice supporting S-parameter analysis.

- [x] Add INDQ and CAPQ devices #143
- [ ] Add microstrip lines models
- [x] Add S2P file devices support #233

wsteffe:
From ngspice user manual (section 6.2.1 Lossy Transmission Line Model (LTRA)) I have seen that this simulator supports trnsmission lines and I have also seen that qucs-s provides a circuit element of this kind.

The only problem is that input data (O=LTRL specification) is not user friendly. I would expand it. 
In example:

---Input data
  Z=
  epr_eff=
  len=
  alpha=(Nepers/length) 

----  created ngspice model:
 //RLC (uniform transmission line with series loss only):
  G=0
  Vp=c/sqrt(epsr_eff)=1/sqrt(LC) 
  C=Z/Vp
  L=Z^2 C
  R=2 alpha Z

wsteffe:
Evren better:

--input:
alpha_dB=(dB/length) = conductorLoss/L computed in Tool/Line Calculator

---- created ngspice model:
R=2 (alpha_dB/20*ln(10)) Z

wsteffe:
May I ask to add also an ideal Coupled Transmission Line such as that one of ADS shown below?.  
It should be very easy to implement it using ngspice Coupled Conductor Line Model (see at https://nmg.gitlab.io/ngspice-manual/transmissionlines/kspicelossytransmissionlines/coupledmulticonductorline_cpl.html)
![ADS_CLIN](https://github.com/ra3xdh/qucs_s/assets/12879321/d4c24f43-b514-4316-bae5-03a970def2d3)

tomhajjar:
Not sure you saw Holger's last ngspice update on 2023-08-04 that includes built in S-parameter file support.

https://sourceforge.net/p/ngspice/discussion/120973/thread/51228e0b01/?page=3

ra3xdh:
[...]

PR #564, opened 2024-02-14
Add RLCG device support for Ngspice

ra3xdh:
This PR makes RLCG device available for all SPICE simulators. This is part of #94


ra3xdh:
image

PR #565, opened 2024-02-14
Reduce view drifting away when using Ctrl + mouse wheel to zoom

wawuwo:
Hi! A small fix to schematic zooming.

This patch doesn't eliminate the drift, but replaces it with something more like a "jitter". This makes zoom-ins and zoom-outs more stable comparing to the old behavior where zoom-ins and zoom-outs made the view to drift towards bottom-right corner.


ra3xdh:

PR #581, opened 2024-02-17
Custom simulation fixes

ivandi69:
This PR addresses a regression I found in netlist rework where the prefix enclosed by # is not applied when there is only a custom simulation in the schematic. It also parses the nutmeg script code to detect the simulations used and avoid unnecessary prefixing. Finally it changes regexes to allow variable names containing points to be detected in the customsim dialog and extracted by scalars print parser.

### DIFF
--- a/qucs/extsimkernels/abstractspicekernel.cpp
+++ b/qucs/extsimkernels/abstractspicekernel.cpp
@@ -1175,12 +1175,14 @@ void AbstractSpiceKernel::convertToQucsData(const QString &qucs_dataset)
         bool hasParSweep = false;
         bool hasDblParSweep = false;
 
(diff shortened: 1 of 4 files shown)


zergud:
In what version do you offer this and #582 PR?
24.1.1?

ivandi69:
The PRs are targeted at current branch, so 24.1.1, I guess.

zergud:
there is **release/24.1** branch for bugfixes only
this has not been announced, because development rules are in the process of change
so, all NEW features in **current** branch
bugfixes only for 24.1 in **release/24.1**

your PR can be merged in current (with squash) and "backported" to 24.1

PR #588, opened 2024-02-18
Make Qt6 building workflow trigger fire only when the code is changed

wawuwo:
Hi! In #586 I completely forgot that there are _two_ workflows and changed only the one with Qt5. This PR does the same but for Qt6 workflow this time.

Citing from #586:
>Adding path filters to the trigger makes it more "granular" and gurantees that building process doesn't start when something that doesn't affect the build (like README or translation) is changed.
>
>see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -2,10 +2,10 @@ name: CMake
 
 on:
   push:
-    branches: [ "master","current" ]
+    branches: [ "master", "current", "release/*" ]
     paths: [ "**.cpp", "**.h", "**/CMakeLists.txt" ]
   pull_request:
-    branches: [ "master","current" ]
+    branches: [ "master", "current", "release/*" ]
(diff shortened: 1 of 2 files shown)


zergud:
may be need to add brances starts with "release/" to both workflow?

wawuwo:
@zergud yes, it sounds reasonable. I've update the branch. This makes PR violate "one logical unit" rule… But on the other hand, guidelines haven't been installed yet, so maybe it's OK for this time :^)

zergud:
Let's not put ourselves in too strict limits )

wawuwo:
@zergud, no, let's make it separate PRs or I won't be able to sleep at night :)

wawuwo:
Ha-ha! I was to late :)

Issue #186, opened 2022-11-22
Parameter sweep issues


tomhajjar:
I know we discussed this but I'm still seeing issues when doing Parameter sweeps of tubes and FETs. I think the issue may the same for any Parameter sweep using an even number of Points. 

In this example below, I want plots of Vgrid=0 -> Vgrid=-max. If I use Start=-5, Stop =0, Step=1, Points=6, the first plot is for Vgs=-0.833 not Vgs=0. 

To get the plot Vgs=0, I have to change Points to 5. For this setting it says Step is 1.25 when Step is actually 1.

![2022-11-21_184906](https://user-images.githubusercontent.com/102838604/203181640-cabed0ae-4d16-4a23-af0b-cf2da988bdac.jpg)

![2022-11-21_184734](https://user-images.githubusercontent.com/102838604/203181630-9dc0b504-e1bc-4d5d-9adf-7f18a4b762d4.jpg)

[Tube_6CA7_Push_Pull_prj.zip](https://github.com/ra3xdh/qucs_s/files/10061395/Tube_6CA7_Push_Pull_prj.zip)

ra3xdh:
The Ngspice and Qucsator has different meaning of "step" and "intervals count" sweep. This issue is explained here #8

tomhajjar:
So Steps is wrong for ngspice and xyce spice but correct for Qucs. Since Qucsator is being depreciated, is it time to make a change or use the selection of simulator to change the definition?

cmat21:
It is somewhat confusing to have a difference in DC sweeps using the same settings. Is it planned to correct this difference?
![dc sweep diference](https://github.com/ra3xdh/qucs_s/assets/70080894/270efb10-00e5-4d6d-98b7-3ea29ab34667)

ra3xdh:
The DC sweep using DC+ParSweep was inherited from old Qucs. Qucsator has no DC sweep simulation and uses this combination instead. It also uses different method to calculate step. The same explantion as here: https://github.com/ra3xdh/qucs_s/issues/8 I will think how to provide a fix and don't break the compatibility between simulators. It is not an obvious task.

ra3xdh:
Fixed by #615

PR #603, opened 2024-02-21
DC simulation variable names and stored values

ivandi69:
This PR makes DC simulation behave as the other simulations.
Variable names are v() for voltages and i() for currents. Only named nodes and probes are stored.
Now DC sweeps and DC simulations are conflicting and prefixing is applied.


ra3xdh:
Thanks for the contribution! Merged. @ivandi69 You may also look at #186 This would fit the Ngspice netlister redesign. Parameter sweep has a point number/step issues. It still be unclear how to provide a solution and keep Qucsator compatibility. Probably you can give some ideas.

PR #627, opened 2024-03-04
Fix #624 (tentative)

### LINKED ISSUE #624: Regression Xyce parametric sweep
I am not able to plot results from Xyce parametric sweeps. This seems a regression because I see from older example file that it was possible in the past.
I made certain experiments with file format std and raw without success.
Bildschirmfoto vom 2024-03-02 16-19-34

[BSIM6_xyce_prj.zip](https://github.com/ra3xdh/qucs_s/files/14469866/BSIM6_xyce_prj.zip)

ivandi69:
This PR should fix #624.
I've never used Xyce and don't have it installed, so only the netlist generation was tested.

### DIFF
--- a/qucs/extsimkernels/xyce.cpp
+++ b/qucs/extsimkernels/xyce.cpp
@@ -134,8 +134,9 @@ void Xyce::createNetlist(QTextStream &stream, int , QStringList &simulations,
 
     //execute simulations
 
-    QFileInfo inf(Sch->DocName);
-    QString basenam = inf.baseName();
+    //QFileInfo inf(Sch->DocName);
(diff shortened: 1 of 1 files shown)


ra3xdh:
I have tried this solution myself. This is not sufficient. The `res_file` variable and res file name is wrong. The `dataset_prefix` variable is not set when using  Xyce. See `abstractspicekernel.cpp` lines 1247-1248.

ra3xdh:
I have missed that you have changed the basename too. I have compiled and tested your approach. Now everything works as expected for Xyce. Thanks for providing fix. Merged.

PR #630, opened 2024-03-07
Add support for '_props.json' generation from osdi file.

onehebe:
The original '_props.json' generator would cause some troubles, especially when too many comments or macros were shown in the verilog-a code (like in cmc-recommend format).
The revised code provide a solution to obtain the correct values for parameter name/value/description from osdi interface.
When osdi file not exits, the json file will be generated following the old routine.


ra3xdh:
Thanks for the contribution! This may resolve problems with JSON reported in #411. I would merge this. But I have no intention to further develop JSON approach for Verilog-A support, because it was designed for ADMS and has many disadvantages. The ADMS development has been discontinued. I have proposed a workflow without JSON files. See: https://github.com/ra3xdh/qucs_s/issues/411#issuecomment-1873331501 I have intention to implement a special file device which could automate this approach. But this is a low priority task for me because of low demand on compact modelling features among Qucs-S users.

Issue #416, opened 2023-12-28
Qucs-S 24.1-7df03e9 -  Yet Another Tuner Slider Bug


3813127458:
Probably this little bug is related to system language (EN) and system locales (DE) combination(?):

[Tuner_Slider_Bug.zip](https://github.com/ra3xdh/qucs_s/files/13783886/Tuner_Slider_Bug.zip)
[Screencast from 2023-12-28 07-06-47.webm](https://github.com/ra3xdh/qucs_s/assets/75557552/b0b7d6cc-f17e-4b3c-b414-ce78ea794253)

ra3xdh:
Yes, it seems the parser cannot recognize the decimal delimiter for DE locale. I will look into this.

ra3xdh:
@3813127458 Could you tell your `locale` settings? I have tried with `de_DE` locale and the behavior is a bit different.

3813127458:
The `locale`  command gives the following listing:      
                                                                                                                                                              
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC=de_DE.UTF-8
LC_TIME=de_DE.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=de_DE.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER=de_DE.UTF-8
LC_NAME=de_DE.UTF-8
LC_ADDRESS=de_DE.UTF-8
LC_TELEPHONE=de_DE.UTF-8
LC_MEASUREMENT=de_DE.UTF-8
LC_IDENTIFICATION=de_DE.UTF-8
LC_ALL=

On the actual development branch Qucs-S 24.1-9fc4db38 the problem still exists.

I use the following desktop:

Operating System: Manjaro Linux 
KDE Plasma Version: 5.27.10
KDE Frameworks Version: 5.113.0
Qt Version: 5.15.11
Kernel Version: 6.6.8-2-MANJARO (64-bit)
Graphics Platform: X11

3813127458:
If you have a fix I will test it.
The tuner slider is a key feature ... should work very solid in the next major release.

We should think also about two further issues (optional):

1) Tuner limits and step size saved optional in parameters `VAL[[[:MIN]:MAX]:STEP]`.
This is useful in many cases, as it is not required to re-define the tuner slider parameters every time again.
Could be used also for optimizer parameter limits.

[...]

PR #636, opened 2024-03-09
Make tuner inputs accept fractional numbers only in "C" format

wawuwo:
Hi! This is a quick fix of #416 based on a solution offered in [this comment](https://github.com/ra3xdh/qucs_s/issues/416#issuecomment-1970618256)

Recap of the problem:
Input fields on tuner form use QDoubleValidators which by default operate in user's native locale. But unfortunately some other parts of tuner codebase are not locale-aware, which leads to bugs if user's native locale number formatting differs from that in "C".

This PR makes tuner inputs unaware about native locale and use "C" instead, i.e. accept fractional numbers only if
 - dot is integral/fractional part separator
 - there is no group separators in number

Scientific notation is still accepted.


ra3xdh:
Thanks for providing fix! Merged.

Issue #567, opened 2024-02-15
Qucs-S user library portability


tomhajjar:
I noticed that the two RF resistor examples "Test_chip_res_basic.sch" and "Test_RATF_EquivCircD.sch" don't work anymore. The devices used in these examples work in the 7 other examples in the attached project. 

https://github.com/ra3xdh/qucs_s/tree/current/examples/ngspice/RFLumpComp

One issue is the examples you uploaded point to your user_lib folder. I didn't dig too deeply into the other issues but it got me thinking how to make user created projects and libraries "portable". LTspice allows placing everything required into a single folder. Is there a way other than making subcircuits containing the symbol information to place a user created Qucs compatible library in the project folder?

**TODO** list

- [x] Add schematic directory to library search paths;
- [x] Resolve relative paths in the library device property dialog
- [x] Make the libraries from the project directory visible in the library manager; Add group *Project libraries*

[Resistor_RF-Models_prj.zip](https://github.com/ra3xdh/qucs_s/files/14298924/Resistor_RF-Models_prj.zip)

ra3xdh:
> Is there a way other than making subcircuits containing the symbol information to place a user created Qucs compatible library in the project folder?

There is no way to place library at the project directory. Library devices support specifying the full path to libraries, but library manager will look only at system library or user_lib directories. This was already reported and related to #4. It still be unclear how to introduce library search paths and improve the library portability.

ra3xdh:
Relative paths implemented with #640. I will implement the visibility of the libraries in project directory myself.

ra3xdh:
The #676 implements showing the libraries from the project directory in the library tree.
![image](https://github.com/ra3xdh/qucs_s/assets/4920080/f9c5299d-a617-4d42-b9ab-b1ccb7505db6)

PR #640, opened 2024-03-13
Add support for relative file paths

ivandi69:
This PR addresses portability issues like #567 by adding support for relative file paths.

File paths are stored as relative paths if they are in the schematic folder or a sub-folder of it.
If they aren't, absolute paths are used.

Resolving paths is done as follows:
If absolute path doesn't exist the filename is searched in the schematic folder, then in the work folder (/home/user/.qucs), then in the Search Path List from Settings->Locations.
A relative paths is searched relative to schematic folder. If it doesn't exist a filename search is done in the work folder, then in the Search Path List.
The search fails silently returning the original path. I am not sure if a non existing file always leads to simulation error.

The PR allows to ship all needed files and libraries in the schematic folder and sub-folders, but it doesn't resolve the library manager issues. A library in the schematic folder is used, but not shown in the library list.


ra3xdh:
I have tested this and didn't find major issues. Are you going to add more commits to this?

ivandi69:
No.

Issue #145, opened 2022-08-26
Use xdg base directory specification


moshiur-raj:
Currently qucs uses ~/.qucs . This contributes to littering the home directory. According to the xdg specification configs should be located at ~/.config , data should be located at ~/.local/share and cache should be located at ~/.cache .
Simply moving ~/.qucs to ~/.local/share/qucs would probably be the easier choice.

ra3xdh:
Yes, I am planning to move `.qucs` elsewhere but with the different purpose. The `$HOME/.qucs` directory doesn't hold config. The temporary files (netlist and log ) are located at `$HOME/.qucs/spice4qucs` (probably could be moved to .cache). The config file location is `$HOME/.config/qucs/qucs_s.conf` The `$HOME/.qucs` is used as the default location for projects and user library file. This is not good, because the hidden direcotory is not visible in file manger for default that may make difficult finding files. 

I propose the following layout for these directories:

*  `$HOME/.config/qucs/qucs_s.conf` config file. Default Qt applications config location.
*  `$HOME/cache/qucs` for holding temporary files like netlist and simulator data
* `$HOME/QucsWorkspace` default location for projects and user libraries. Not a hidden directory because need to visible from file manager by default.

moshiur-raj:
Qucs automatically creates .qucs in my home directory, even when I change the location in settings. Can you fix this too?

ra3xdh:
Yes, Qucs-S now creates default directory layout on the first start now. It is related to fix of #86. The usage of `QStandardPaths` instead of hardcoded paths may help to resolve this issue https://doc.qt.io/qt-5/qstandardpaths.html

ra3xdh:
The implementation of this provided by #658

moshiur-raj:
qucs still creates ~/.qucs at startup @ra3xdh .

PR #658, opened 2024-03-23
Change Qucs-S workspace and tempfile paths

wawuwo:
Hi!

This pull request addresses three issues described in #145.

1. Set root path for all temporary files produced by Qucs-S to QStandardPaths::CacheLocation (which is `~/.cache/qucs-s` on GNU/Linux)
2. Set default Qucs-S workspace path to `~/QucsWorkspace`
3.  Fix a bug which led to recreation of default workspace dir on app start even if it was changed in app settings

I've tested all three changes, but testing of the first one was incomplete: I've tested only the `ngspice` simulation because I don't have a single clue how to run other types like VHDL, ASCO, etc. (I am complete noob in electrical engineering to be honest)

### DIFF
--- a/qucs/components/opt_sim.cpp
+++ b/qucs/components/opt_sim.cpp
@@ -80,7 +80,7 @@ QString Optimize_Sim::netlist()
 bool Optimize_Sim::createASCOFiles()
 {
   Property* pp;
-  QFile afile(QucsSettings.QucsHomeDir.filePath("asco_netlist.cfg"));
+  QFile afile(QucsSettings.tempFilesDir.filePath("asco_netlist.cfg"));
   if(afile.open(QIODevice::WriteOnly)) {
     QTextStream stream(&afile);
     stream << "*\n";
@@ -164,7 +164,7 @@ bool Optimize_Sim::createASCOFiles()
     afile.close();
   } else return false;
 
-  QDir ExtractDir(QucsSettings.QucsHomeDir);
+  QDir ExtractDir(QucsSettings.tempFilesDir);
   if(!ExtractDir.cd("extract")) {
     if(!ExtractDir.mkdir("extract"))
       return false;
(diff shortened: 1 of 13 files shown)


ra3xdh:
Thanks. I will test this and share the feedback. It's need to use fresh installation to test this. To test VHDL/Verilog you should install `ghdl` and `iverilog` and make it available from $PATH. Then simulate schematic like in this thread: https://github.com/ra3xdh/qucs_s/issues/97#issuecomment-1562399016 Digital simulation starts automatically if the schematic contains `Digigtal` simulation type. I didn't use ASCO for a long time and I am not sure if it works :smile: ASCO works only with qucsator.

wawuwo:
@ra3xdh I've found a bug: directory for simulator output (which is defined in "Simulator settings") is always set to `~/.cache/qucs-s/spice4qucs` at app start. In other words custom path for simulator output is respected until app termination only.

I'll prepare a fix to respect user-defined setting after restart. But I'm in doubt: if simulator output is temporary data and temporary data should be stored in *standard* location, then shouldn't there be no way to set custom simulator output dir?

wawuwo:
>I think there is no need to set the simulator output directory by user after switching to standard paths

I see; then I'll adjust simulator settings dialog and revert last fix 🤝

ra3xdh:
I have tested this and found one major issue. The new Qucs-S always sets the project directory to `$HOME/QucsWorkspace`. The application shown an empty project tree and user libraries list after restart. Users need to move their data manually. This may lead to tons of false positive bugreports after releasing this feature. The following logic should be used here:

* Don't change projects tree location, if the old project directory `QucsSetting.HomeDir` (usually points to `$HOME/.qucs`) contains some projects or user libraries. Probably need to check for `_prj` subdirectories (projects) and `user_lib` subdirectory.
* If there is no contents in `$HOME/.qucs` then migrate to `$HOME/QucsWorkspace`
* Ideally ask user if the data should be migrated and copy contents of `$HOME/.qucs` to `$HOME/QucsWorkspace`

Probably you can propose a better solution. I will merge this PR after this issue will be resolved.

ra3xdh:
Also the solution may be to use new workspace location only on the fresh installation of the application. The `QucsSettings.firstStart` could be used.

wawuwo:
@ra3xdh, oh, I see :(
That's because I decided to use another name for config property, "QucsHomeDir" property containing old path is ignored. I'll try to come with a fix today

wawuwo:
@ra3xdh I implemented it the easy way: if it is a fresh install (no `~/.config/qucs/qucs_s.conf` or no property in config), then set `~/QucsWorkspace`, otherwise respect the value from config file.

ra3xdh:
Thanks! Now everything works correct. Merging.

PR #659, opened 2024-03-24
Refactor Schematic::sizeOfAll

wawuwo:
Hi!

This PR rewrites wordy parts of `Schematic::sizeOfAll` body in more concise manner. Function semantics and interface are not changed.


ra3xdh:
Merged. Please keep in mind that `Components` etc. are not `QList` but `Q3PtrList` It may have some difference with QList. I am wondering that range iterating works on it as expected. In the long perspective Q3PtrList must be replaced by QList everywhere.

PR #673, opened 2024-04-05
Add MTEEs to microstrip quarterwave filter

jcolafrancesco:
This commit is bringing a fix to quarterwave microstrip filter generation so they render properly within the RF layout visualizer. Previously, wires were connected to more than 3 components and this was causing the visualizer to fail. Those wires have been replaced by Tees.

Sadly, it was not possible to do an incremental fix without refactoring the quarterwave_filter class.


ra3xdh:

PR #674, opened 2024-04-06
Use standard open-file dialog to open example schematics

wawuwo:
With this commit the click on "Examples" entry of "File" menu instead of opening a new file manager window starts a standard dialog to select and open a file.

I reused logic from `QucsApp::slotFileOpen` and altered it a bit: example files don't go to "recent files" list.

Also a handful of new `tr` strings are introduced which lack translations. I guess I'll make a separate PR after running `lupdate` on all files, removing obsolete entries too.

Closes: ra3xdh#672

### DIFF
--- a/qucs/qucs.cpp
+++ b/qucs/qucs.cpp
@@ -1962,12 +1962,19 @@ bool QucsApp::closeAllFiles()
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


ra3xdh:

PR #697, opened 2024-04-20
Fix tabdiagram headers with Qucsator data

andresmmera:
It has been noticed that the data headers are missing from the Tab Diagram when the simulation data comes from Qucsator instead of Ngspice.

### Data from Qucsator [NOT OK]
imagen

### Data from Ngspice [OK]
imagen

By inspecting the code in "tabdiagram.cpp", It was found that line [250](https://github.com/ra3xdh/qucs_s/blob/current/qucs/diagrams/tabdiagram.cpp#L250C1-L250C34) filters the name of the dependent variable by removing all characters before the "/". 

The data coming from a Ngspice simulation has names like "ngspice/variable_name", but in the case of Qucsator, the "/" is missing, so the table headers are empty.

### Dataset from the Ngspice simulation
imagen

### Dataset from the Qucsator simulation
imagen

This PR fixes the above problem by first checking that the string contains "/":

### After this fix
imagen

A workspace is provided to verify this:
[TESTBENCH_TABLES_wo_headers_if_qucsator_prj.zip](https://github.com/ra3xdh/qucs_s/files/15048786/TESTBENCH_TABLES_wo_headers_if_qucsator_prj.zip)

### DIFF
--- a/qucs/diagrams/tabdiagram.cpp
+++ b/qucs/diagrams/tabdiagram.cpp
@@ -247,7 +247,9 @@ int TabDiagram::calcDiagram()
     y = y2-tHeight-5;
     colWidth = 0;
 
-    Str = g->Var.section('/', 1);
+    Str = g->Var;
+    if (Str.contains('/')) Str = g->Var.section('/', 1);
+
     colWidth = checkColumnWidth(Str, metrics, colWidth, x, y2);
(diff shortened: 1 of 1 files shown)


ra3xdh:
Thanks for providing the fix for this bug! Merged.

Issue #679, opened 2024-04-10
Improve SpiceLibComp device


ra3xdh:
SpiceLibComp device was intended for usage of the unmodified SPICE libraries with Qucs-S https://qucs-s-help.readthedocs.io/en/latest/SubLib.html#usage-of-unmodified-spice-libraries The current implementation has many issues that prevents its usage for circuit design. The following improvements must be done:

- [x] Provide more symbol file patterns in `share/qucs-s/symbols`
- [x] Define the syntax of *.sym files. Probably a common XML schematic header need to be added to these files
- [x] Provide a simple way for symbol editing: #147 #700 
- [x] Recreate 4 existing SYM files in `share/qucs-s/symbols` subdirectory using new format. See #700 See #762
- [x] Provide a simple way to assign the SUBCKT pins and symbol pins without wrapping symbol in Qucs subcircuit #755
- [x] Implement *New symbol* toolbar button and action #757 
- [x] Fix the `nested subciruit` warning that appears with Ngspice and may cause simulation convergence issues

Help is appreciated here.

wawuwo:
@ra3xdh I've made a small patch enabling Qucs-S to save only a subcircuit symbol of a schematic to a file and to load back symbols from such files. Patch reuses `Schematic` object and resulting "symbol files" are in effect stripped usual schematic files (no components, no properties etc.)

Workflow to create a symbol:
1. Create new schematic 
2. Add the required number of ports
3. Switch to symbol editing mode
4. Draw symbol
5. Save symbol by going to `File->Save` as and selecting "Symbol only" type of file

From now on one can load this symbol back using usual "File->Open" dialog, where "Symbol only" type of file has to be selected.

Unfortunately with this patch symbols from `share/qucs-s/symbols` can't be loaded — they lack `<Qucs Schematc … >` tag on first line.

Please check out branch https://github.com/wawuwo/qucs_s/tree/feature-export-symbol in my repo if it sounds interesting, I think it's not worth of opening of a PR yet.

wawuwo:
I forgot to mention that SpiceLibComp can use symbol files made this way, but you have to manually put them into `share/qucs-s/symbols`

ra3xdh:
> I've made a small patch enabling Qucs-S to save only a subcircuit symbol of a schematic to a file

Yes, this may help to resolve #147 I was thinking about the similar way. The main problem was how to add/remove ports for symbol without schematic. I will try to evaluate your patch in the next week.

[...]

PR #700, opened 2024-04-22
Add symbol editor

### LINKED ISSUE #147: Add editor for symbol `*.sym` files
Any reason "Qucs" never had a Symbol editor? I have been "stealing" symbol sections out of other peoples schematics using a text editor or use symbols.sch from QucsStudio. Editing Qucs-S schematics using a text editor is not user friendly. 

On a side note, I complained how frustrating it was to make good looking transformers and coils using Elliptic Arc on the Qucs SourceForge forum. I was told to use a text editor to alter the Elliptic Arc parameters :-(.  I always give up and steal an inductor from symbols.sch which I got from QucsStudio.

https://sourceforge.net/p/qucs/discussion/311050/thread/b833231fb8/

[symbols.zip](https://github.com/ra3xdh/qucs_s/files/9456341/symbols.zip)

wawuwo:
Opening a PR following the discussion in #679

Recap:
- Symbol editor based on usual schematic editor
- Saving and opening detached symbols to and from *.sym files
- To save and open detached symbols, "Symbol only" filter should be selected in "Open file" dialog
- To start creating a symbol, switch to symbol editing mode
- Ports can be added and removed in symbol editing mode. If the symbol is then saved as *.sym file then all manually inserted ports remain. If it is saved as usual *.sch file, then all ports without a corresponding port in schematic will be removed.

May close #147

### DIFF
--- a/qucs/paintings/portsymbol.cpp
+++ b/qucs/paintings/portsymbol.cpp
@@ -53,7 +53,7 @@ void PortSymbol::paint(ViewPainter *p)
   p->Painter->setPen(QPen(Qt::red,1));  // like open node
   p->drawEllipse(cx-4, cy-4, 8, 8);
 
-  QSize r = p->Painter->fontMetrics().size(0, nameStr);
+  QSize r = p->Painter->fontMetrics().size(0, nameStr.isEmpty() ? numberStr : nameStr);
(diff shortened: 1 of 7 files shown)


ra3xdh:

PR #708, opened 2024-04-26
Refactor Component::getTextSelected

wawuwo:
Hi!

This one refactors the function responsible for determining which text property of a component is clicked:
- obscure `Corr` arg indirectly coming from `Schematic::textCorr` is removed
- function body is rewritten in a more clear and easy to understand way
- function is documented
- semantics remain the same. 

If something is affected by these changes that would be the "click-on-component-property" behaviour; I checked it before pushing and it looked like it was before.

### DIFF
--- a/qucs/components/component.cpp
+++ b/qucs/components/component.cpp
@@ -202,6 +202,7 @@ int Component::getTextSelected(int point_x, int point_y) {
 
     for (auto* prop : Props) {
         if (!prop->display) {
+            text_index += 1;
             continue;
         }
 
--- a/qucs/components/component.cpp
+++ b/qucs/components/component.cpp
@@ -142,37 +142,79 @@ void Component::getCenter(int &x, int &y) {
     y = cy;
 }
 
-// -------------------------------------------------------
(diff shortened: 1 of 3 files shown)


ra3xdh:
I have tested this PR. Something works wrong with components having  long properties list like library transformer device. Look at the attached screencast. The invisible properties are selected. I am using Qt6.7.0+Wayland. 
[Screencast_20240523_091320.webm](https://github.com/ra3xdh/qucs_s/assets/4920080/a3cc6682-af86-4fba-9d69-bbf70aff7a28)

wawuwo:
@ra3xdh I've fixed it, that was my mistake: while iterating over component's properties the index number describing current property wasn't increased if the property was hidden. It made the index number "lack behind" and point to a wrong property. The bug was not related to a long properties lists, it could be reproduced with any component: one just had to make a "gap" in properties (i.e. "visible, visible, invisible, visible, visible") and click on any property after the gap. 

I hope it'll be OK now! 😎

ra3xdh:
Thanks for the contribution! Now everything works as expected.

PR #718, opened 2024-05-13
added windows.yml

dsm:
Hi, this PR will add basic MSYS2 CI for three different variant ucrt64,mingw64 and clang64

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,67 @@
+name: windows
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


ra3xdh:
Thanks! I am merging this. The next step is to find a way to prepare at least portable package and save build artifact. This will allow to provide snapshots for Windows users.

dsm:
Can you create an issue for this and we discuss how to deploy portable package which step we need

ra3xdh:
See #719

PR #721, opened 2024-05-16
Turned off the output of component parameters in the schemes by default

s4kkkk:
By default, when adding components to the diagram, most of their parameters are output, which greatly clutters the diagram. This can be disabled in the settings of each component, but it is extremely inconvenient. I made it so that all text parameters are not output by default.
image
image

### DIFF
--- a/qucs/components/ac_sim.cpp
+++ b/qucs/components/ac_sim.cpp
@@ -50,13 +50,13 @@ AC_Sim::AC_Sim()
   Name  = "AC";
 
   // The index of the first 4 properties must not changed. Used in recreate().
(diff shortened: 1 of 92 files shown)


ra3xdh:
Thanks for the interest in contributing to Qucs-S, but I cannot accept this PR. Some set of the device properties must always be visible by default. Otherwise the invisible default properties like R,C,L may mislead the user. Some proprietary CAD systems like Cadence Vistuoso also show the device properties by default. This policy was accepted for Qucs since 2003 and is not a subject of change. Please make a proposal at the Discussion/Ideas  https://github.com/ra3xdh/qucs_s/discussions before making such massive PRs next time.

PR #725, opened 2024-05-22
added windows ci continuous release

dsm:
Hi,

this PR add snapshot release for windows

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -120,6 +120,7 @@ jobs:
         path: ~/artifacts
 
     - name: Check if continuous_build release exists
+      if: github.event_name == 'push'
       run: |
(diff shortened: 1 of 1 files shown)


ra3xdh:
Thanks! The snapshot build seems to work. What is the purpose to include DX shader compiler DLLs in the snapshot package?

dsm:
Because of windeployqt-qt6.exe shows warning to not found them

ra3xdh:
The DXC compiler DLLs seems to be optional dependency. I have tried to remove this from distribution and everything works fine. I have tried both Linux/Wine and physical Windows 11 platforms.  The warning from `windeployqt` may be ignored.

ra3xdh:
Thanks for the contribution! I have merged this.

dsm:
For 2 weeks I have master exams and then I will make improvements.

PR #728, opened 2024-05-31
Fix rendering quirks of some lumped components

wawuwo:
Hi! This is a set of small changes to some of lumped components to deal with rendering artifacts which became visible after switching to new schematic rendering.


ra3xdh:

PR #729, opened 2024-05-31
Rearrange includes

wawuwo:
Hi! This PR removes some unused or already transitively included headers and replaces some "umbrella" headers with more specific ones. `qucs-s` is only affected. Nothing special, but a bit of tiding things up, everything stays the same. On my machine this reduces the time of `make qucs-s` from 20 to 15 minutes.


ra3xdh:
Yes, the includes optimization would be useful. I will merge this after MINGW build issues will be resolved.

PR #735, opened 2024-06-04
Fix source components rendering quirks

wawuwo:
Hi!

This a follow-up to #723 and another portion of changes to components to make them look a bit prettier. This time components from "sources" groups were in work.

With this PR also comes a new type of drawing primitive — the so called "Polyline" which allows to draw compound lines consisting of several segments and make good-looking joining of them. I had to add be able to draw nice arrow heads without strange edges etc.

There is a lot of affected components and I was to lazy (shame on me!) to make before/after images this time, but I think a change in component's rendering is something self-describing and easy to see so lack of before/after images shouldn't be a problem.

### DIFF
--- a/qucs/components/am_modulator.cpp
+++ b/qucs/components/am_modulator.cpp
@@ -24,15 +24,19 @@ AM_Modulator::AM_Modulator()
   Description = QObject::tr("ac voltage source with amplitude modulator");
   Simulator = spicecompat::simQucsator;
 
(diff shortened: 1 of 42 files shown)


wawuwo:
Oops, turns out I've confused directions of a couple of arrows. Preparing a `--fixup` commit…

ra3xdh:

PR #740, opened 2024-06-07
Fix #739: Restore QPainter state after paining the id_text element

### LINKED ISSUE #739: Selection Box, Selecting Items and Moving Items Issues
The selection box has radically changed making it unusable on Windows. The mouse Cursor is not even close to the corner of the selection box. Why is the selection box using thick lines? It makes selecting items in close proximity to others difficult.

It's almost impossible to select items as well. The cursor has to be placed up to the right away from the item you want to select. Trying to select a single item when multiple items are nearby is impossible.

Also when I select an item and try to move it, it jumps down to the left away from the cursor making moving or resizing items impossible. 

It's literally impossible to make or edit a schematic, sub-circuit or model. I'm trying to test the new Varactor lib but had to give up. Everything I try to edit turns into garbage. The old method used in Qucs is preferred.

wawuwo:
The 'paint' function of ID_text missed the restore() call on painter, which led to carrying the changes made to painter further down the call chain. In other words, the painter state *after* painting the ID_text wasn't the same as *before* doing it, and everything painted after the ID_text was affected by this change in painter's state.

This commit add missing restore() call.

Fixes: ra3xdh/qucs_s#739

### DIFF
--- a/qucs/paintings/id_text.cpp
+++ b/qucs/paintings/id_text.cpp
@@ -63,6 +63,7 @@ void ID_Text::paint(QPainter* painter) {
     painter->setPen(QPen(Qt::darkGray,3));
     painter->drawRoundedRect(-4, -4, x2+8, y2+8, 4.0, 4.0);
   }
+  painter->restore();
 }
 
 // --------------------------------------------------------------------------


ra3xdh:
Thanks! Now everything works as expected on my side. Waiting for feedback from @tomhajjar

ra3xdh:
Thanks for the fix! Merging.

PR #741, opened 2024-06-08
Make border lines of selection rectangle be thin

wawuwo:
The change to selection rect we discussed earlier in #739


ra3xdh:
Thanks for providing fix. Merging!

Issue #750, opened 2024-06-11
Issues Editing Sub-circuit Symbols


tomhajjar:
Still having issues editing and creating Sub-circuit Symbols. #739 didn't fix all the issues.

The small thin black "+" at 0,0 disappears during editing and gets replaced by a larger thick black "-". I then lose track of the "Centroid" for the symbol. I have to type Esc to get it to reappear but then I can't edit. It happens if you select any item(s) like a line and move it.

The Port labels ANODE/CATHODE are randomly disappearing, I can only get the the labels back if I go back to Schematic mode and then re-enter Symbol editing mode. Using "Undo" doesn't work. This happens in symbol editing mode if you select the entire diode symbol and then use "CTRL R".

In symbol editing mode if you select the entire diode symbol, move it and then use "CTRL R", the entire symbol disappears and "Undo" does nothing. I have to go back to Schematic mode and then re-enter Symbol editing mode. 

If I select an item and type "CTRL R" the item disappears. Doesn't matter if in schematic or symbol editing modes. "CTRL R" only works if I type it first and then select the item I want rotated. This is not the way it used to work.

If I type "CTRL R", Qucs-S is then "Stuck" in Rotate mode and I have to type "ESC". 

**TODO** list

- [x] fix coordinate center rendering in symbol mode #752 
- [x] fix undo operation in symbol editing mode #754 
- [x] disable rotate shortcut while moving graphics primitives in symbol editing mode 

![2024-06-11_144857](https://github.com/ra3xdh/qucs_s/assets/102838604/546669ca-b038-4508-822a-d9b6b98cc89a)

![2024-06-11_150241](https://github.com/ra3xdh/qucs_s/assets/102838604/838c0180-1263-49e2-8fdb-99ab43a3fa3d)

![2024-06-11_150907](https://github.com/ra3xdh/qucs_s/assets/102838604/9cd460e8-b2f9-4aea-9a84-82d1550f5519)

[SOD323_pkg.zip](https://github.com/user-attachments/files/15793656/SOD323_pkg.zip)

wawuwo:
Hi! 

I've managed to reproduce two bugs and I'll try to provide fixes for them in the near future.

#### 1. A dash appears at 0,0 when anything on schematic is moved
  Steps to reproduce:
  1. Select any component from left panel and move cursor around schematic or using the mouse move any element already present on schematic

A small dash appears near the coordinates origin cross

[...]

PR #752, opened 2024-06-12
Remove dash at (0,0) which appears when smth on schematic is moved

wawuwo:
Fixes one of the problems described in #750


ra3xdh:
Thanks for the fix! Merged.

PR #757, opened 2024-06-15
Add New symbol action

ra3xdh:
This PR contains the following improvements:

* Added *New symbol* action on toolbar as discussed in #679 
* Implemented `Schematic::isSymbolOnly` property.
* Locked switching schematic/symbol if document is symbol only. 
* Suggest the `sym` extension on *Save As* if document is symbol only. Not possible to save symbol as schematic. Saving schematic as symbol only works.
* Added new `symnew.svg` icon


ra3xdh:
I have added the following fixes: 
* The `PortSymbol::numberStr` must be always be a number. Otherwise it may cause symbol load issues. 
* Allow delete port in symbol only mode

PR #762, opened 2024-06-17
Rework SpiceLibComp symbols

ra3xdh:
This PR contains the following changes:

* The unsupported `AnalogueCM` and `MESFETs` libraries has been deleted
* The SpiceLibComp symbols moved from `qucs/spicecomponents/symbols` to `library/symbols`
* The 4 existing symbols updated using new symbol file format. Now it is possible to open symbols using the symbol editor.


ra3xdh:
The extended symbols pack will follow later.

PR #763, opened 2024-06-17
Qucs Filter - Fix quarter wavelength coupled filter implementation

andresmmera:
I have found that the QW transmission line filter synthesis was not working properly after commit [601548](https://github.com/ra3xdh/qucs_s/commit/6015488fc24c746aad6ebec5991c51e26d65be2b). Specifically, the impedance of the shunt resonators was set to Z0 regardless the design parameters. This leads to excessively wide passbands.

image

In addition to this, I have found that when the MTEE was added in commit [601548, L124](https://github.com/ra3xdh/qucs_s/commit/6015488fc24c746aad6ebec5991c51e26d65be2b#diff-88ec292e540c7ae794efecb233cf45be5ed1f23b41b9f1b1785448e9425316e8R124), the length of the coupling was shortened in an attempt to compensate for the extra length added by the tee. As far as I know, this does not restore the original filter response. Notice that the MTEE equivalent circuit adds stray circuit elements that degrade the filter response so the compensation seems not to be as simple as that.

I couldn't find any literature on how to compensate for this analytically, so IMHO it's better to manually tune the couplings and resonator variables.

Please have a look at the attached workspace [3] and the PR notes [4]. They contain 4 examples of QW coupled resonator BP and BS filters. These examples cover the complete filter design and (except for the 4-8 GHz BSF case) they show good agreement between the ideal transmission line filter, simulation with MS models and EM data obtained using [Sonnet Lite](https://www.sonnetsoftware.com/products/lite/).

image

**Other issues**

A user has pointed out to me (via private email) that the length of the microstrip resonator was calculated using the dielectric permittivity instead of the effective dielectric permittivity. This caused a frequency shift in the design. This issue is addressed in this PR.

**Other changes**

- It was found duplicity on the transmission line filter and MS implementation logic [601548, L113](https://github.com/ra3xdh/qucs_s/commit/6015488fc24c746aad6ebec5991c51e26d65be2b#diff-88ec292e540c7ae794efecb233cf45be5ed1f23b41b9f1b1785448e9425316e8L113). This issue is addressed in this PR.
- Added MS open to BSF open stubs. The MS open circuits account for the fringing effects and omitting this leads to a frequency shift in the filter response. This was not included in the original code.
- Added MS vias to terminate the shunt resonators in the BPF implementation. This is more realistic than having an ideal ground.
- The physical values of the filter have been rounded to two decimal places. This reduces clutter in the schematic and is more than enough in practice.

**References**

[1] [Tian, Ge et al. “Research on the equivalent circuit of microstrip T-junctions and its application on branch-line coupler.”
2012 International Conference on Microwave and Millimeter Wave Technology (ICMMT) 1 (2012): 1-4.](https://ieeexplore.ieee.org/document/6229911)
 
[2] Microstrip tee junction. Qucs Manual. https://qucs.sourceforge.net/tech/node81.html

[3] PR test workspace. [TEST_QW_FILTER_prj.zip](https://github.com/user-attachments/files/15873718/TEST_QW_FILTER_prj.zip)
[4] [Spreadsheet with commit details and tests. ](https://docs.google.com/spreadsheets/d/1K1G0WspB7u91xZWsOQpYXJdZXGEHzE0MlGIpDEbM_uw/edit?usp=sharing)

### DIFF
--- a/qucs-filter/filter.cpp
+++ b/qucs-filter/filter.cpp
@@ -415,6 +415,30 @@ QString Filter::getMS_Via(double height, int x, int y, int rotate)
     return code;
 }
 
+QString Filter::getMS_Open(double width, int x, int y, int rotate)
+{
+    QString code;
+    switch (rotate)
+    {
+        case 0: // No rotation
+            code = QString("<MOPEN MS93 1 %1 %2 -20 -50 0 0 \"Sub1\" 0 \"%3 mm\" 1 \"Hammerstad\" 0 \"Kirschning\" 0 \"Kirschning\" 0>\n").arg(x).arg(y).arg(QString::number(width*1e3, 'f', 2));
+            break;
+
+        case 1: // CTRL+R
+            code = QString("<MOPEN MS93 1 %1 %2 15 -12 0 1 \"Sub1\" 0 \"%3 mm\" 1 \"Hammerstad\" 0 \"Kirschning\" 0 \"Kirschning\" 0>\n").arg(x).arg(y).arg(QString::number(width*1e3, 'f', 2));
+            break;
+
+        case 2: // 2 x (CTRL+R)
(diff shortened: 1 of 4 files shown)


ra3xdh:

PR #777, opened 2024-06-20
Generalise Text::angle() logic

### LINKED ISSUE #765: Sub-circuit Text Displayed
Sub-circuit symbol text is not displaying properly when the sub is placed on a schematic. The text looks correct when creating the symbol but wrong when symbol placed in the schematic. Text is rotated 90 deg and is wrong font or size.

[Photo_Voltaic_Isolators_prj.zip](https://github.com/user-attachments/files/15878987/Photo_Voltaic_Isolators_prj.zip)

wawuwo:
Hi!

I've rewritten `Text::angle()` to make it generic and able to handle any combination of `mCos` and `mSin` values. Before that `Text::angle()` could handle only three distinct cases and all other cases fell in "default", it was rather "stiff". I believe that is the reason why subcircuit symbol has text with wrong rotation angle.

It looks good now (at least on my machine):
image

May fix #765

### DIFF
--- a/qucs/element.cpp
+++ b/qucs/element.cpp
@@ -18,6 +18,7 @@
 #include "element.h"
 #include "misc.h"
 
+#include <cmath>
 #include <QPainter>
 
 namespace qucs {
--- a/qucs/element.cpp
+++ b/qucs/element.cpp
@@ -70,28 +70,37 @@ void Text::draw(QPainter *painter, QRectF* br) const {
 }
 
 double Text::angle() const {
-  // Historically Text uses mSin and mCos values to store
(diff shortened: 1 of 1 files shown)


dsm:
Hi, why you don't use `std::atan2` ? 

It gives the exact angle and handle all quadrants.

wawuwo:
@dsm thanks for the advice! I've rewritten it using the suggested function and the code looks much better and simpler now. The very existence of the tangent function had completely slipped out of my mind somehow and I chose that straightforward approach.

ra3xdh:
Thanks for the fix! Merged.

PR #780, opened 2024-06-23
Fix Pac to match result from Ngspice and Qucsator

### LINKED ISSUE #779: What is the rationale for the 0 dBm voltage value in power source?
### Discussed in https://github.com/ra3xdh/qucs_s/discussions/778

<div type='discussions-op-text'>

<sup>Originally posted by **csrabak** June 22, 2024</sup>
The magnitude of the voltage source for the Qucs-s power source, available in the components->sources pane puts 0.316228 V as magnitude for a 0 dBm power, 50 Ω (default impedance).

The line of the exported netlist is:

```
VP1 out 0 dc 0 ac 0.316228 SIN(0 0.316228 1MEG) portnum 1 z0 50
```

Since simulators use magnitude for this parameter, which is the _peak_ voltage of the source, this value would be OK for an open circuit, but due the internal impedance only one fourth of a milliwatt is available to the matched load.

Shouldn't the voltage be such that when a matched (i.e. 50 Ω) load is fed the power delivered to the load be 0 dBm (1 mW)?</div>

ra3xdh:
This PR fixes #779 and #778 I have corrected the equations to match result from Ngspice and Qucsator.

### DIFF
--- a/qucs/components/source_ac.cpp
+++ b/qucs/components/source_ac.cpp
@@ -106,7 +106,7 @@ QString Source_ac::ngspice_netlist()
     double z0 = spicecompat::normalize_value(getProperty("Z")->Value).toDouble();
     double p = spicecompat::normalize_value(getProperty("P")->Value).toDouble();
     double vrms = sqrt(z0/1000.0)*pow(10, p/20.0);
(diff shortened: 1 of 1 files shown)


ra3xdh:
image

Issue #730, opened 2024-05-31
Wrong bound box for simulation block


zergud:
does not take into account text size in other languages
so, texts overlap and selection rectangle wrong

![image](https://github.com/ra3xdh/qucs_s/assets/869819/84d8b97e-877a-409f-879a-45ca64bd05d8)

wawuwo:
@zergud assign this on me please, this is definitely the consequence of rendering refactoring

PR #783, opened 2024-06-23
Refactor simulation components

wawuwo:
Hi!

I started working on #730 and found out that in order to fix it it's better to refine components inheritance hierarchy a bit, so I've come up with this. A short description of changes:
1. Distinction between drawing of simulation and usual components
```
if (Model.at(0) == '.') {   // is simulation component (dc, ac, ...)
    drawSimulator(p);
} else {    // normal components go here
    drawUsual(p);
}
```
is removed from `Component` in favour of a single method — `Component::drawSymbol`. By default it draws a usual component and anyone wanting their own logic has to override it.

2. A new class `SimulationComponent` is added: it inherits from `Component` overriding its `drawSymbol` to provide logic specific to simulation components. This class is meant to be a common base class for all simulation components.
3. All simulation components are changed to inherit from newly created base class .

There are a lot of commits, but most of them are tiny, and the most important ones are the first two. They lay the foundation for all subsequent ones.

### DIFF
--- a/qucs/components/CMakeLists.txt
+++ b/qucs/components/CMakeLists.txt
@@ -61,6 +61,7 @@ circline.cpp
 taperedline.cpp
 circularloop.cpp
 spiralinductor.cpp
+simulation.cpp
 )
 
 SET(COMPONENTS_HDRS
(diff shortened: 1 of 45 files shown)


wawuwo:
There is a problem: with some fonts simulation "box" has wrong size; working on it

wawuwo:
Now it's OK I hope

ra3xdh:
I didn't find issues with this PR. Merging. Thanks for the contribution!

PR #810, opened 2024-07-05
fixed git hash.

dsm:
Hi, this PR fix this.

I tried mac built-in git 2.39.5 and it gives no warning and no hash,
but recent version installed via homebrew gives warning
also windows and linux do the same thing.

- git gives a warning -> fatal: '1u': not an integer


ra3xdh:
Thanks! Merged. Another problem is that the latest Windows continuous build has the `share` subdirectory missing. It contains only `bin` subdirectory inside the archive.

dsm:
I'll fix this thanks.

dsm:
Same naming typo in zip the files not ./re  it sould be ./share

Issue #818, opened 2024-07-07
Dark theme and rendering issue on windows


dsm:
Hi, Qt added new theming for win11 named Windows11 style beginning Qt 6.7. This theme can automatically switch dark and light mode. In dark mode editing component color of the value is white. And some rendering artifact. I add some pictures. @wawuwo

Problem 1:

![image](https://github.com/ra3xdh/qucs_s/assets/41967334/abcef82a-f2ae-4dba-a19d-2e14fc0052a3)

Problem 2:

![image](https://github.com/ra3xdh/qucs_s/assets/41967334/9a08c6e2-1851-4ba3-9e92-85063ab9018f)

Problem 3:

![image](https://github.com/ra3xdh/qucs_s/assets/41967334/da1e7e7f-fa5b-4fc4-bbd3-30d46f005f07)

edit from @ra3xdh : Added **TODO** list

- [x] When property value is being edited, value text is misplaced
- [x] When property value is being edited and QT theme is switched to dark mode, value text has white color and is barely visible
- [x] When property value is being edited, value text is visible only partially
- [x] Connections to component ports are rendered incorrectly; if port has only one connection is has to be rendered as a little filled square

ra3xdh:
I have also recently noticed that the dots on device pins are disappeared after schematic rendering refactoring. 
* Old 
![image](https://github.com/ra3xdh/qucs_s/assets/4920080/f3f336b4-6cc3-48cd-9ca6-2b1d0b6222bb)
* New -- no dots
![image](https://github.com/ra3xdh/qucs_s/assets/4920080/bff09661-a3c1-4a9c-a903-6ece668bf648)

wawuwo:
Hi guys!

Here is a checklist of problems to check that I've understood them well and to control the progress. Please correct me if I'm wrong:

- [ ] When property value is being edited, value text is misplaced
- [ ] When property value is being edited and QT theme is switched to dark mode, value text has white color and is barely visible
- [ ] When property value is being edited, value text may is visible only partially
- [ ] Connections to component ports are rendered incorrectly; if port has only one connection is has to be rendered as a little filled square

dsm:
The application is already in dark theme, but when the value is changed, there is a white background and white text in windows11, but in macOS, this does not happen in the dark theme, it is white text on a black background. In problem 2, there is also a residue of the value that normally appears on the white background, which should not be there.

ra3xdh:
[...]

PR #828, opened 2024-07-08
Use stylesheet to change appearance of QLineEdit used for editing a property value

wawuwo:
In ra3xdh/qucs_s#818 @dsm mentioned that when property value text is being edited, the color of text background depends on current colorscheme (dark/light), which sometimes makes text barely visible.

With this commit the background color of QLineEdit, used to modifying property value, is set using stylesheet instead of QPalette.


dsm:
I tested background issue is fixed, thanks.

ra3xdh:
Thanks! Merged.

PR #838, opened 2024-07-12
Enable variables for blue nonlinear devices

ra3xdh:
This PR contains the following fixes:

* Enabled variables for blue devices parameters
* Rename *Qucs legacy equation* to *Qucsator equation*
* Deprecate SweepModel property of Parameter sweep simulation. See #839 
* Fixed chopping of Qucsator variable names in diagram axis labels

Here is the test case for the fix. The version before fix gives error on this schematic and doesn't accept `Bf=BF1`.

image

### DIFF
--- a/qucs/components/component.cpp
+++ b/qucs/components/component.cpp
@@ -739,9 +739,7 @@ QString Component::form_spice_param_list(QStringList &ignore_list, QStringList &
             } else {
                 nam = Props.at(i)->Name;
             }
-            double val, fac;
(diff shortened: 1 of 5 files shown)


tomhajjar:
I redid the BJT examples with this fix. I used the latest Qucs-S nightly and the new ngspice 43

[2N2222_Amplifiers_ngspice_prj.zip](https://github.com/user-attachments/files/16241673/2N2222_Amplifiers_ngspice_prj.zip)

PR #842, opened 2024-07-14
Fix property editbox position and size

wawuwo:
Hi!

This is an attempt to fix problems with property value edit box, which were discussed in #818.

It works pretty accurate, though edit box location is not 100% precise, there are tiny fluctuations. Editing of property value is implemented by placing borderless QLineEdit on top of property, it's quite hard to compute the exact location for it, which won't cause text to "jump" a bit.


ra3xdh:
Thanks! Everything works as expected on my side. Merging.

PR #843, opened 2024-07-14
Fix ASCO running with QucsatorRF

### LINKED ISSUE #830: ASCO optimization doesn't work with QucsatorRF but works with Qucsator
The reason for this issue is that `qucsator` executable name is hardcoded inside ASCO sources. Also ASCO always assumes that `qucsator` is available from $PATH. This may make difficulties for running a portable version. The ways to resolve this:

* Provide a patched version of ASCO with the next release
* Submit a patch to ASCO that allows to set the path to qucsator excutable

ra3xdh:
This is an attempt to fix #830. The idea is to create a symlink `qucsator_rf->qucsator` in system temporary files directory.


ivandi69:
The way the symlink is created is a security risk. https://cwe.mitre.org/data/definitions/61.html

> On POSIX systems, the path may be the one specified in the environment variables TMPDIR, TMP, TEMP, TEMPDIR, and, if none of them are specified, the path "/tmp" is returned.

It would be better to use the .cache folder, or use https://en.cppreference.com/w/cpp/io/c/tmpnam

PR #844, opened 2024-07-14
Get rid of MouseActions::drawn

wawuwo:
Hi!

Probably you've seen lines like
```
drawn = false;
```
or
```
view->drawn = false;
```
and so on scattered all around the codebase. This flag is mostly set to 'false' and only occasionally to 'true', and when it's 'true' it enables blocks of code which are described as "erase old symbol". I honestly tried to grasp its meaning and purpose, what "old symbol" is being erased, and I failed. I started thinking that it would be easier to remove it and see what effect it would make. 

I tried removing it and surprisingly (or not?) there was no difference at all. Everything looked and behaved as usual, at least at first sight. And now I am asking you to kindly include this change to mainline to test it in real life conditions (testing in prod, yeah 😞).

If there will be no problems discovered in, say, week or month, than we'll be happy having get rid of the dead code. In other case I'll revert it. I crammed everything in a single large commit for the sake of easier reverting if it will be needed later.

### DIFF
--- a/qucs/mouseactions.cpp
+++ b/qucs/mouseactions.cpp
@@ -178,7 +178,6 @@ void MouseActions::editLabel(Schematic *Doc, WireLabel *pl)
 
     Doc->updateAllBoundingRect();
     Doc->viewport()->update();
-    drawn = false;
     Doc->setChanged(true, true);
 }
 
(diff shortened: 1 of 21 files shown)


ra3xdh:
I can schedule the merging of this soon after the v24.3.0 release. I am planning to prepare the release packages in this week. I have assigned v24.4.0 milestone to this PR.

ra3xdh:
I didn't find any side effects on this PR. Merging.

PR #832, opened 2024-07-09
some refactoring.

dsm:
Hi, this PR do some refactoring.

Issue #748, opened 2024-06-11
Get rid of Q3PtrList wrapper


ra3xdh:
This is a long term refactoring task. It doesn't directly affect the simulation or program appearance, but will improve the code readability after implementation. Currently the main application uses a wrapper for `Q3PtrList` class in `qt3_compat/q3ptrlist.h` It would be good to replace this with `QList`.

**TODO** list
- [x] Remove Q3PtrList in `Node` class #845 #881 
- [x] Remove Q3PtrList in diagram dialog #854
- [x] Port `Component::Properties` to QList #858 
- [ ] Port remaining members of Schematic class to QList or std::list (???)

wawuwo:
Getting rid of Q3PtrList has a subtle complication: some parts of the codebase use the "auto delete" feature. 
```
qucs % grep -rF setAutoDelete | grep -vF qt3_compat
./schematic.cpp:    DocComps.setAutoDelete(true);
./schematic.cpp:    DocWires.setAutoDelete(true);
./schematic.cpp:    DocNodes.setAutoDelete(true);
./schematic.cpp:    DocDiags.setAutoDelete(true);
./schematic.cpp:    DocPaints.setAutoDelete(true);
./schematic.cpp:    SymbolPaints.setAutoDelete(true);
./schematic.cpp:    Wires->setAutoDelete(false);
./schematic.cpp:    Components->setAutoDelete(false);
./schematic.cpp:    Wires->setAutoDelete(true);
./schematic.cpp:    Components->setAutoDelete(true);
./schematic.cpp:    Wires->setAutoDelete(false);
./schematic.cpp:    Components->setAutoDelete(false);
./schematic.cpp:    Wires->setAutoDelete(true);
./schematic.cpp:    Components->setAutoDelete(true);
./schematic.cpp:    Wires->setAutoDelete(false);
./schematic.cpp:    Components->setAutoDelete(false);
./schematic.cpp:    Wires->setAutoDelete(true);
./schematic.cpp:    Components->setAutoDelete(true);
./schematic.cpp:    Components->setAutoDelete(false);
./schematic.cpp:    Components->setAutoDelete(true);
./schematic.cpp:    Wires->setAutoDelete(false);
./schematic.cpp:    Wires->setAutoDelete(true);
./mouseactions.cpp:        Doc->Wires->setAutoDelete(false);
./mouseactions.cpp:        Doc->Wires->setAutoDelete(true);
[...]

PR #845, opened 2024-07-16
Refactor node

wawuwo:
Hi!

I've refactored the Node class to address the problem from #748.

@dsm seems to be working on the same issue in #832, but in a broader scope, I hope this piece doesn't interfere with his work. Anyway some synchronisation may be required upon merging (if it happens).

List of changes:
- Remove dependency on Q3PtrList (most important)
- Hide implementation details (in particular make private the container used to actually store connected elements)
- Completely new API
- Formatting and code structuring
- Move 'Conductor' from element.h to a separate header file (not related to Node, just sorting things out)

I've done some tests on my side, in particular:
- lay horizontal wire, lay vertical wire
- lay two-segment (L, Г, etc.) wire
- lay multiple-segment wire, i.e. chain of wires (click, move cursor and click, move cursor and click, and so on.)
- connect wire to a component
- start new wire from a middle of existing wire
- end new wire in a middle of existing wire 
- move connected component around
- delete connected wire segment, delete connected component

All described above went smoothly, but one cannot be 100% sure; some nasty bug is probably hiding waiting for perfect conditions to occur :(

### DIFF
--- a/qucs/CMakeLists.txt
+++ b/qucs/CMakeLists.txt
@@ -132,6 +132,7 @@ SET(QUCS_SRCS
 
 SET(QUCS_HDRS
 element.h
+conductor.h
 main.h
 messagedock.h
 misc.h
--- a/qucs/conductor.h
+++ b/qucs/conductor.h
@@ -1,0 +1,17 @@
+
+#ifndef CONDUCTOR_H
+#define CONDUCTOR_H
+#include "element.h"
+
+class WireLabel;
+
+/** \class Conductor
+  * \brief label for Node and Wire classes
+  *
+  */
+class Conductor : public Element {
+public:
+  WireLabel *Label;
+};
+
+#endif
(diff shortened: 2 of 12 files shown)


dsm:
Hi, some changes related but my work not done properly, I just got back from holiday, I choose this way for Connections in node.h  `std::list<std::weak_ptr<Element> > Connections;` we have too many raw pointer in qucs-s so we should use smart pointers for refactoring. We need to refactor thousands of lines of code :(

in sweepdialog.cpp I removed this lines not confusing for loop variables:   
```
Node *pn;
Element *pe;
```

ra3xdh:
I have tested this and didn't found issues with schematic editing, save, load. I am merging this.

wawuwo:
@ra3xdh,I am almost certain that the reason is that connected elements are being removed from node's internal `std::list` in [deleteWire](https://github.com/ra3xdh/qucs_s/blob/2d1a86d127ecfa9911742e4406c772dae1ce3aa0/qucs/schematic_element.cpp#L2571) while iteration over this list (indirectly, through Node's `begin()` and `end()`) is in progress in [insertComponentNodes](https://github.com/ra3xdh/qucs_s/blob/2d1a86d127ecfa9911742e4406c772dae1ce3aa0/qucs/schematic_element.cpp#L2558), and it breaks the iterator.

I don't have a solution right now, need some time to think what could be done

Issue #852, opened 2024-07-20
Fresh Qucs-S installation tries to set working directory to system root /


ra3xdh:
This old bug has been fixed, but it seems to appear again after recent refactoring. Steps to reproduce:

* Move existing `$HOME/.config/qucs/qucs_s.conf` to a safe place or run Qucs-S in a fresh user profile
* Try to simulate schematic and you will get the following error:
 
This error blocks the upcoming release. 

![image](https://github.com/user-attachments/assets/6d42284b-6629-44bb-97dc-4a5e6568b107)

ra3xdh:
Also the toolbar is not visible after the fresh start. 

![image](https://github.com/user-attachments/assets/93948a8a-fb8d-461c-9239-4f6f36653a25)

ra3xdh:
I will provide a fix for this soon.

ra3xdh:
Fixed by #853.

PR #853, opened 2024-07-21
Fix Qucs-S fresh start #852

ra3xdh:
This PR provides the fix of the application fresh start #852 and contains some refactoring.

* Fixed the working directories paths after the application fresh start #852 
* The toolbars and status bar are always visible now. The related settings removed. 
* Removed dark/light icon theme setting from application settings dialog. This setting has no effect because since auto-generated SVG icons are used since v2.1.0 The remaining controls only mislead users.


ra3xdh:
Merged. The fresh application start has been fixed on all platforms.

ra3xdh:
In the long perspective it will be needed to get rid of old QucsSettings API everywhere.

PR #854, opened 2024-07-22
Replace Q3PtrList in diagram dialog

wawuwo:
Hi!

This is another piece for #748

### DIFF
--- a/qucs/diagrams/diagramdialog.cpp
+++ b/qucs/diagrams/diagramdialog.cpp
@@ -110,7 +110,6 @@ DiagramDialog::DiagramDialog(Diagram *d, QWidget *parent, Graph *currentGraph)
 {
     setAttribute(Qt::WA_DeleteOnClose);
   Diag = d;
-  Graphs.setAutoDelete(true);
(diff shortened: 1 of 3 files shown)


dsm:
Hi, can we use QList instead of std::vector for this refactoring.

wawuwo:
@dsm hi! Is there a specific reason for it? `QList` and `std::vector` both array-based, `Graph` is a private member of diagram dialog and is not passed as argument to any of QT APIs

dsm:
Btw I Quick test QList over std::vector soke compiler error happend, I think QList wants to copy uniquie ptr somewhere I don't but unique ptr doesn't copyable just movable maybe we can use weak_ptr or shared_ptr instead of unique if there isn't spesific desicion.

wawuwo:
I am more inclined to keep the `std::vector` for this one… It works nice and easy in all cases except of removing an element, and here is only one case of removing (plus I just like to stick with standard library facilities when possible). But anyway, thanks for the advice and attention to details

ra3xdh:
Merged. I didn't find issues while testing this.

PR #858, opened 2024-07-22
Migrate Properties to QList and Some Refactoring

dsm:
Hi, This PR migrate Properties to QList #748 , getFirst and getLast method name in Q3PtrList to front and back so QList, std::vector etc. has front and back so this changes will reduce refactoring effort.

we should test to ensure stability of this PR.


ra3xdh:
I will merge this short after the upcoming release.

dsm:
this PR became a Properties migration to QList.

dsm:
there is bugs somewhere I don't find. I tested bjt noise example in latest snapshot build no problem but this PR release gives this in ngspice console I manually stop in macos because normally it run 1-2 sec.  

`Ngspice started...
alter=0: no such command available in ngspice
alter=0.10001: no such command available in ngspice
alter=0.20002: no such command available in ngspice
alter=0.30003: no such command available in ngspice
alter=0.40004: no such command available in ngspice
alter=0.50005: no such command available in ngspice
alter=0.60006: no such command available in ngspice
alter=0.70007: no such command available in ngspice
alter=0.80008: no such command available in ngspice
alter=0.90009: no such command available in ngspice`

ra3xdh:
The BJT-noise runs as expected on my machine. Do you mean this example?

image

ra3xdh:
> I tested bjt noise example in latest snapshot build no problem but this PR release 

I will look at this. I am planning to start to review the refactoring PRs in a few days after the release. Is only BJT-noise example failing, or all parameter sweep simulations affected?

ra3xdh:
> there is bugs somewhere I don't find. 

Probably something in `Param_Sweep::getNgspiceBeforeSim` need to be rewritten to align with refactored properties. Also you may check the `Component::getProperty`

dsm:
if I open BJT.sch and BJT_swp.sch app crashed normally don't crash. somewhere in the code broken

dsm:
Bugs in component:load method for BJT.sch

ra3xdh:
Some devices (at least Diode) use custom `load` procedure for backward compatibility. Something like this may be applied for BJT too. I will look into this in the next few days.

dsm:
I found a fix for crash loading BTJ.sch for equations section call Props.insert method and QList sometimes moves List to another memory location and this happend p1 iterator null and not valid below code is the solution.

```C++
  if (Props.size() < (counts >> 1)) {
    int index = std::distance(Props.begin(), p1);
    Props.insert(index + 1, new Property("y", "1", true));
    p1 = Props.begin() + index;
  }
```

dsm:
finally i fixed parameter sweep.

dsm:
How I do this?

ra3xdh:
Save the both netlists from before/after and use some software like `kdiff`. It will show what entry is different.

dsm:
24.3
```
XTRAN4 0  0 _net6 _net2 _net3 Transformers_PositiveCouplingPS K=0.99 L1=1 L2=1 Rp=1 Rs=1
XTRAN5 0  _net8 _net7 IF 0 Transformers_PositiveCouplingPS K=0.99 L1=1 L2=1 Rp=1 Rs=1
```

this PR
```
XTRAN4 0  0 _net6 _net2 _net3 Transformers_PositiveCouplingPS p=0.99 K=1 L1=1 L2=1 Rp=1
XTRAN5 0  _net8 _net7 IF 0 Transformers_PositiveCouplingPS p=0.99 K=1 L1=1 L2=1 Rp=1
```

ra3xdh:
Yes, I see now that the last property of Transformer is not loaded. I guess all library devices may be affected.

dsm:
p=0.99 is it correct I think not

ra3xdh:
Yes, the property name `K` is not loaded. It seems the property names are shifted by one position.

dsm:
latest commit fixed the issue.

image

dsm:
@ra3xdh Hi, I recently push 2 commits. fixed build natively Qt 6.8 beta 2 with MSVC(VLAs not supported) and refactoring cmake file to swtich c++20 and only enable warning for debug build.

ra3xdh:
> refactoring cmake file to switch c++20 

Be careful with the latest C++ extensions. The Ubuntu-22.04 is not EOL and need to be supported till 2027. This distribution may not support some C++20 features. Need to check if this compiles and works using Ubuntu-22.04. But I have recently dropped support for Ubunutu-20.04 which is EOL next year. 

> Btw MSVC build not ok.

MSVC support may be dropped. If this will work with MSYS2+GCC/Clang this will be sufficient.

dsm:
ubuntu 22.04 has gcc-11.4 and it support c++20 and can be install gcc-12.3. Maybe I will test MSVC Clang(Clang-cl).

ra3xdh:
I have recently merged #845. @dsm, You may consider making the rebase of this PR.

dsm:
I lost my recent changes being rebase please wait :)

dsm:
no conflict with #854

ra3xdh:
I have tested this PR with my simulations and didn't find any visible issues. Are you planning to add more commits to this? I am intending to test this in the next few days including creating schematics, subcircuits, and libraries from scratch and then merge this if no issues will be found.

dsm:
Hi, @ra3xdh I have no plan to commit this PR if you found any issue, I'll try to provide solution.

dsm:
I'll provide final fix for MSVC build .

dsm:
MSVC build not fixed properly sorry, debugger says App->view null if I click and select component pane then click schematic then crash. I have no solution for this. I'll test older version because normally it worked maybe I can find which update broke.

dsm:
@ra3xdh Hi, I test oldest version for MSVC crash and after version 2.1.0 MSVC build not working and crashed same as 24.3.

dsm:
@ra3xdh Hi, I test oldest version for MSVC crash and after version 2.1.0 MSVC build not working and crashed same as 24.3.

I found which commit cause this bug. after #397 MSVC build broken.

ra3xdh:
> I found which commit cause this bug. After https://github.com/ra3xdh/qucs_s/pull/397 MSVC build broken.

If it is unclear how to fix, the MSVC support fix may be postponed. The MSYS2 support is sufficient.

dsm:
I look what difference so many formatting changes so I manually reformat old and new files and look diff using winmerge I don't find the issue so we should postponed may be fixed later. this PR has served its purpose, let's not waste any more time.

dsm:
I found the solution no code changed only compiling option needed if you compile via /vmg everythings works.

https://bugreports.qt.io/browse/QTBUG-90735?focusedId=551355&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-551355

dsm:
@ra3xdh you found any issue ?

ra3xdh:
Yes, I have found a problem. The application crashes when trying to insert ground regardless from the button on toolbar or from the left panel. The copy-paste of the ground also doesn't work. But the schematics are opened without problems and simulation works. The insertion of other 1-port devices like  *Subcircuit port* works as expected. The application crashes only on attempt to insert ground.

ra3xdh:
Here is the possible fix to the ground insertion issue:
~~~
diff --git a/qucs/schematic_element.cpp b/qucs/schematic_element.cpp
index 4b6a2a93..3e73142e 100644
--- a/qucs/schematic_element.cpp
+++ b/qucs/schematic_element.cpp
@@ -2505,6 +2505,7 @@ bool Schematic::distributeVertical()
 // digital sources and sets them accordingly.
 void Schematic::setComponentNumber(Component *c)
 {
+    if (c->Props.isEmpty()) return;
     Property *pp = c->Props.front();
     if(!pp) return;
     if(pp->Name != "Num") return;
~~~

ra3xdh:
I have found another problem. The editing properties of the equation type devices works wrong. How to reproduce. 

* Place the *Nutmeg equation* on schematic.
* Add 3-4 equations.
* Press Apply button and application crashes at `componentdialog.cpp` line 1065

The same may appear with *.PARAM* section. Application doesn't crash but one equation is dropped. Sometimes application may crash at `componentdialog.cpp` line 1089.

dsm:
I solved this issue append will changed address same as the insert.

dsm:
Latest commit solved issue can you test vhdlfile and optimizedialog ?

ra3xdh:
The optimization doesn't work as expected. The dialog doesn't crash, but the existing properties cannot be loaded. You can open the `examples/qucsator/filter_optimization.sch` as the test schematic. 

* Correct
image
* This PR
image

dsm:
Now works.

image

ra3xdh:
I have found another crash. Place on schematic a simulation having simulation dialog like AC or TRAN, then change something and press apply. Application crashes in `Component::slotApplyInput`.

dsm:
I sent two new commit.

ra3xdh:
The crash has gone, but the problem with simulation still present. The `Points` property is duplicated for TRAN after `Apply` or `OK` is applied. The `MaxStep` property is chopped. Probably the whole `Component::slotApplyInput` needs redesign using new APIs. Currently it still use iterating style inherited from Qt3. 

image

ra3xdh:
> Yes it needs to be redesign some times it has to remove somestimes not

I will try to provide a redesigned version of `ComponentDialog::slotApplyInput` in the next few days. The properties should be removed from this dialog only for equation-type components. Other devices should not allow to delete properties.

dsm:
I will wait for your new design

dsm:
I remove latest 2 commit and rebase with current.

ra3xdh:
I have finished with redesign of the `ComponentDialog` See https://github.com/dsm/qucs_s/pull/1 The crashes seems to be gone now.

dsm:
it seems everythings works fine after your PR.

ra3xdh:
I have found another crash while testing this PR. See https://github.com/ra3xdh/qucs_s/pull/845#issuecomment-2262938687 for context. The most probably it is related to #845

dsm:
yes latest snapshot also has this issue and so it's not related this PR.

ra3xdh:
@dsm, Please apply the following patch on `subcircuit.cpp` Otherwise netlister adds `File` as the first parameter and simulation fails. 

~~~
diff --git a/qucs/components/subcircuit.cpp b/qucs/components/subcircuit.cpp
index a8207c1e..61541112 100644
--- a/qucs/components/subcircuit.cpp
+++ b/qucs/components/subcircuit.cpp
@@ -230,8 +230,9 @@ QString Subcircuit::spice_netlist(bool)
         s += " "+nam;   // node names
     }
     s += " " + misc::properName(f);
-    for(Property *pp : Props) {
-        s += QString(" %1=%2").arg(pp->Name).arg(spicecompat::normalize_value(pp->Value));
+    for(int i = 1; i < Props.size(); i++) {
+      s += QString(" %1=%2").arg(Props.at(i)->Name)
+               .arg(spicecompat::normalize_value(Props.at(i)->Value));
     }
     s += "\n";
     return s;
~~~

ra3xdh:
Yes, Qucsator shows warning on `File` property, but the simulation works as expected. The `Subcircuit::netlist` also must be fixed.

ra3xdh (approved):

ra3xdh:
I didn't find any further problems with this PR and I am merging this. @dsm, Please stay in connection if something related will appear.

Issue #875, opened 2024-07-28
Select-all does not select wires on the boundary


hvwyl:
Select-all operation won't select wires on the boundary

![Image_1722179532524](https://github.com/user-attachments/assets/30ad4d5b-b582-4afc-be81-4008eacf693b)

wawuwo:
@hvwyl, hi! What version of Qucs-S do you use? Do you remember by any chance the last version without this bug? I am able to reproduce the bug in head-of-the-branch build, just need some more info to investigate the cause

hvwyl:
> @hvwyl, hi! What version of Qucs-S do you use? Do you remember by any chance the last version without this bug? I am able to reproduce the bug in head-of-the-branch build, just need some more info to investigate the cause

I have fixed it by changing the properties of `QMargins` in `qucs_actions.cpp`. Although this may not the best solution.

![1](https://github.com/user-attachments/assets/fc817d4a-94a3-482e-8599-bc242bd25db2)

hvwyl:
> @hvwyl, hi! What version of Qucs-S do you use? Do you remember by any chance the last version without this bug? I am able to reproduce the bug in head-of-the-branch build, just need some more info to investigate the cause

I have fixed it by changing the properties of `QMargins` in `qucs_actions.cpp`. Although this may not the best solution.

![1](https://github.com/user-attachments/assets/fc817d4a-94a3-482e-8599-bc242bd25db2)

wawuwo:
@hvwyl I've found the exact place where this bug was introduced, it's the 8c584fc4. In that commit I used `QRect.bottom()` and `QRect.right()` to pass bounds of selection rectangle, but these methods return a value which is smaller than the actual one by one unit ([here](https://doc.qt.io/qt-6/qrect.html#bottom) and [here](https://doc.qt.io/qt-6/qrect.html#right)).

I've opened PR #878, it doesn't *directly* address this bug, but should fix the problem (at least it did while I was testing it).

ra3xdh:
Fixed by #878

PR #878, opened 2024-07-31
Refactor Schematic::selectElements

wawuwo:
Hi!

A major refactoring of one particular member function, this also solves the problem from #875


ra3xdh (approved):

ra3xdh:
The selection works as expected. I didn't find any side effects of this PR.

PR #879, opened 2024-07-31
Remove Schematic::textCorr

wawuwo:
Hi!

Not a great change, just a bit tidying up.

The `Schematic::textCorr` is called only in one place, and the variable which receives its return value is later reassigned new value without being used, i.e. function return value is discarded. All this renders Schematic::textCorr useless and safe to remove.


ra3xdh:
Merged.

PR #881, opened 2024-08-02
Fix segfault when component is placed over a wire

wawuwo:
Hi! This is a fix of the problem we discussed previously in #845


ra3xdh:
Thanks! Merged.

Issue #884, opened 2024-08-08
right scroll issue


sergeiandreyev:
Version: qucs s24.3.0 (12ddd12b)
OS: Ubuntu 22.04 LTS

When I scroll to the right in schematic window Qucs-S crashes with:
```
qucs-s: /home/andreev/opensource/qucs_s/qucs/schematic.cpp:2463: void Schematic::scrollRight(int): Assertion `step >= 0' failed.
Aborted (core dumped)
```

wawuwo:
@sergeiandreyev, hi! I've prepared a fix for the problem

ra3xdh:
I didn't understand how to reproduce this, but I have accepted #886 as the fix. Fixed by #886. Closing as completed.

sergeiandreyev:
Thank you, @ra3xdh!

PR #886, opened 2024-08-09
Fix assertion fail when schematic is scrolled right by 'Right' key

wawuwo:
Fixes: ra3xdh/qucs_s#884


ra3xdh:
Thanks for the fix! Merged.

PR #894, opened 2024-08-17
add macos qt5 build for older macos version.

dsm:
Hi, this PR add macos qt5 build for older macos version.


ra3xdh:
Thanks, merged! I have some questions regarding Qt5 for MacOS. How long these MacOS versions need to be supported? Which Qt5 version is used by Mac? I am planning to deprecate Qt5 support in the next year, because it is was needed only for Ubuntu-20.04 which is EOL next year. At least I am planning to deprecate Qt versions <5.15.

dsm:
Now it uses latest Qt 5.15.13 i, it needs to build  at least mac os version 10.14, one year maybe enough for support and Qt 6.7 support macos 11 we can set this as minimum.

Issue #871, opened 2024-07-26
add CI support for stable release.


dsm:
Hi, I'm planning to improve our CI to make release more consistent. I added below todo list for this issue. If any steps needs that I missed, please tell me.

- [x] add windows installer support our CI. #903 
- [x] add misc folder to repo. #901
- [x] add inno setup file to repo. #901
- [x] add ngspice to windows release. #903 
- [x] CI should be automatically run for newly created tag and make release. #911 #917 
- [x] insert automatic changelog comparing previous tag to release for maybe NEWS.md not ready. #919

dsm:
@ra3xdh can you add misc folder and inno setup file to repo ?

ra3xdh:
See #901 I have uploaded install scripts in `contrib` subdirectory.

dsm:
@ra3xdh Hi, you build deb and rpm using obs we can add this to our CI to your help, there are github actions for this and runs on ubuntu I added links for them. also we can run freebsd using this action.
https://github.com/marketplace/actions/freebsd-vm

https://github.com/marketplace/actions/build-debian-packages
https://github.com/marketplace/actions/rpm-build

ra3xdh:
It's need to learn the features of Github Actions approach. OBS provides not only DEB/RPM packages build, but also repository infrastructure and automatic generation of installation instructions. Also OBS allows to build the packages for a number of distribution versions. I am not inclined to switch the DEB/RPM generation to Github actions.

zergud:
I suggest changing the versioning for CI, may be 24.3.99
same version as Release may be confusing
![image](https://github.com/user-attachments/assets/01136dfd-8717-423a-b7c9-f1d37d10ef30)

dsm:
@zergud yes, I can do this, it's easy. 24.3.99 or next release ex. 24.4.0 maybe and maybe added git hash. @ra3xdh what is your opinion ?

ra3xdh:
Adding the `99` at the end of continuous release may have side effects. The application checks all three digits in the libraries and schematic. And the schematic or library created using continuous release will show warning in the next stable release for example if releasing `v24.3.1`. It's need to modify the document version check first. Probably comparing only major version would be sufficient.

dsm:
Hi, @ra3xdh I'll send a PR for this, I set nightly version to 23.4.99 and normal release use tag version and hide git hash if it's stable release. I don't use VERSION file for this just used local builds. Now version set is flexible.

[...]

PR #903, opened 2024-08-23
add windows setup support to ci.

dsm:
Hi, @ra3xdh this PR add windows setup in CI to solve #871

- add ngspice to windows portable.
- make curl and wget more robust because of unexpected error break ci.
- use aqtinstall for appimage to build latest qt.
- switch gcc-12 on ubuntu it support more c++20 features.


ra3xdh:
Thanks! Merging.

Issue #905, opened 2024-08-23
Matching Circuit Tool generates wrong results


tomhajjar:
Update: 

The device seems to be marginally stable.
Project updated with all fixes

Follow up to #401, @andresmmera

In testing Matching Circuit Tool for an "L-section match" case, I found the component values to be unrealistic. 

The input match L1 shunt inductor value of "0.000H" should be interpreted as a short to ground or possibly using "creative license", an open circuit to ground. The output match L4 series inductor value of "2.556H" can only be interpreted as an open circuit.

I didn't test any other cases since having to copy and paste S-parameter values becomes tedious and error prone.

![2024-08-23_184159](https://github.com/user-attachments/assets/3b0bd9ec-73d0-49d8-bca8-3d950cad60e9)
![2024-08-23_184019](https://github.com/user-attachments/assets/fd55dba1-162b-4acf-b52b-e2f0748bf8bb)
![2024-08-23_183704](https://github.com/user-attachments/assets/56aebefd-2680-4629-bc2c-9fffbd491cf4)

[BFP193W_amplifier_prj.zip](https://github.com/user-attachments/files/16737223/BFP193W_amplifier_prj.zip)

andresmmera:
Hi @tomhajjar , thanks for pointing this out.

The device is not unconditionally stable and it cannot be conjugately matched. Equations 12.40a/b in [1] don't have a solution for this case since the argument of the square roots is negative.

![image](https://github.com/user-attachments/assets/ed93bb27-91d1-4078-b867-46e7d58b7d6c)

![image](https://github.com/user-attachments/assets/7f548612-48b4-4dfd-bbb5-b228842b1556)

I think that the best solution would be to check if the device is unconditionally stable [here](https://github.com/ra3xdh/qucs_s/blob/current/qucs/dialogs/matchdialog.cpp#L976) in order to avoid wrong results like this. I can write a fix for this later.

Schematic: 
[BFP193W_spar_analysis.zip](https://github.com/user-attachments/files/16735773/BFP193W_spar_analysis.zip)

[1] Microwave Engineering. David M Pozar. 4th Edition. 2012

tomhajjar:
Not often a device's Spice model correlates so well with it's S-parameter data...

![2024-08-24_085057](https://github.com/user-attachments/assets/54d4b3f2-c573-47e3-acc2-c27fab906e24)
[...]

PR #907, opened 2024-08-24
Fix issue #905 - The matching network tool gives incorrect results when conditionally stable data is provided

andresmmera:
As reported in issue #905, the matching network tool gave incorrect  results when using a conditionally stable S-matrix.To obtain a conjugate match at both ports, the device must satisfy the unconditional stability criteria, defined as K > 1 and abs(delta) < 1. See [1].

A condition has been added [here](https://github.com/andresmmera/qucs_s/blob/Fix_Issue_905_Matching_Tool_K_less_1/qucs/dialogs/matchdialog.cpp#L745C1-L745C33). It checks the stability and stops the execution if the criteria is not met.

### 1) Conditionally stable device (#905)

Using the S-parameter data provided in #905, a message is now displayed warning of conditional stability and advising the user to consider adding losses and/or FB.

Selection_019
Selection_020

### 2) Unconditionally stable device.

The device in 1) is stabilized with 12 Ohm at the input. This makes it unconditionally stable (K = 1.06 < 1, |delta| = 0.228 < 1)
Selection_021

The tool is used to synthesise the input and output matching networks. It works as usual: Note that S11 and S22 are matched to 50 ohms at the given frequency (400 MHz).

Selection_022
Selection_023

Thank you @tomhajjar

[1] Microwave Engineering. David M Pozar. 2012. Eqs 12.28 and 12.29

### DIFF
--- a/qucs/dialogs/matchdialog.cpp
+++ b/qucs/dialogs/matchdialog.cpp
@@ -732,10 +732,32 @@ void MatchDialog::slotButtCreate() {
                      S12imag * S21imag;
     double DetImag = S11real * S22imag + S11imag * S22real - S12real * S21imag -
                      S12imag * S21real;
-    success =
-        calc2PortMatch(S11real, S11imag, S22real, S22imag, DetReal, DetImag, Z1,
-                       Z2, Freq, micro_syn, SP_block, open_short, Substrate,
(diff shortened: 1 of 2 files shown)


ra3xdh:
Tanks for the contribution! Merged.

PR #909, opened 2024-08-24
Matching tool: Keep dialog open in case the synthesis goes wrong

andresmmera:
As pointed out in issue #905, the dialog closes if the synthesis goes wrong. This can be annoying since the user needs to input all the design data again. Keeping the dialog open allows the user to make corrections.

A condition has been added at the end of the slotButtCreate() function to keep the dialog open in case the synthesis goes wrong. If the synthesis is ok, the program will exit as usual.


ra3xdh:
Thanks! Merged.

PR #925, opened 2024-09-03
Improve temperature sweep

ra3xdh:
## The state of problem

If the schematic contains blue semiconductor devices (D,BJT,JFET,MOS), the SPICE simulation takes the device temperature from `Temp` property and ignores the SPICE global temperature. This makes temperature sweep with SPICE not a straightforward task. It is required to set `Temp=TEMPER` for every device. The explicit setting of `Temp` property is needed for Qucsator compatibility. 

## The solution

I have added the property `UseGlobTemp=yes/no` The default is `yes`. If set to `yes` it uses global temperature. Otherwise the device uses temperature defined in `Temp`. The temperature sweep will be possible for device using default parameters. 

image

### DIFF
--- a/qucs/components/bjt.cpp
+++ b/qucs/components/bjt.cpp
@@ -55,13 +55,17 @@ QString BJT::spice_netlist(bool)
     }
 
     QStringList spice_incompat,spice_tr;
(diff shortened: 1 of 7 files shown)


tomhajjar:
To clarify how to use new feature I created 3 examples. 2 for ngspice and 1 for Qucsator




[BJT_Temp_Sweep.zip](https://github.com/user-attachments/files/16913816/BJT_Temp_Sweep.zip)

PR #991, opened 2024-10-10
Spar viewer: Fixes and New Features

andresmmera:
Hi there.

I've been using the tool and found some issues. While I was at it, I added some features:

### Fixed issues

- Fixed several warnings
- I've noticed that before this PR the marker data was taken from the displayed traces and not from the dataset. This caused the marker data to be clipped when the trace was clipped by the chart limits.
- Before this PR, the tool considers that everything after the first "." is the file extension. Now the tool can handle files with several "."

### New Features

- **Lock button**: A button has been added to freeze the axis settings. This is necessary because the tool automatically adjusts the axes when the user adds a new trace. This behavior is correct and makes the tool more dynamic, but it can be annoying when the user is focused on a specific band and wants to add and remove traces.

- **Limit line:** Most network analyzers I have worked with have this feature. This is used to check if your measurement meets the gain mask, RL, etc.

- **Save / Open Session**: This feature saves the current state of the tool to an XML file. This saves time as the user does not have to enter the limits and marker stuff every time he/she uses the tool.

Here is a screencast:

[https://drive.google.com/file/d/1ZdGebMBIh8tHjbdxuHGOIi8Ydwcz4yNz/view?usp=sharing](https://drive.google.com/file/d/1ZdGebMBIh8tHjbdxuHGOIi8Ydwcz4yNz/view?usp=sharing)

### DIFF
--- a/qucs-s-spar-viewer/main.cpp
+++ b/qucs-s-spar-viewer/main.cpp
@@ -108,7 +108,7 @@ int main( int argc, char ** argv )
   qucs->move(QucsSettings.x, QucsSettings.y);  // position before "show" !!!
   qucs->show();
 
-  QScreen* primaryScreen = QGuiApplication::screens().first();
+  QScreen* primaryScreen = QGuiApplication::screens().constFirst();
 
   qucs->resize(primaryScreen->availableGeometry().size() * 0.9);
   qucs->setGeometry(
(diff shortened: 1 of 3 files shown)


tomhajjar:
Will smith chart and polar plotting be added?

andresmmera:
I would like the tool to be able to plot data on a Smith chart, at least.

I'll try to add that as soon as I can. First I need to find an open source library with a Smith Chart widget, Qt doesn't include that. Otherwise, I'll need to write it from scratch and that'll take more time.

ra3xdh:
Thanks! Merged.

PR #995, opened 2024-10-16
Fix missing text in schematic after upgrading to Qt-6.8.0

ivandi69:
After upgrading to Qt-6.8 calling [QPainter::drawText](https://doc.qt.io/qt-6/qpainter.html#drawText-6) with zero width, zero height and Qt::TextDontClip flag doesn't draw any text.
image
Setting width and height to 1 fixes it.
image

### DIFF
--- a/qucs/components/component.cpp
+++ b/qucs/components/component.cpp
@@ -252,7 +252,7 @@ void Component::paint(QPainter *p) {
     QRect text_br{tx, ty, 0, 0};
 
     if (showName) {
-        p->drawText(tx, ty, 0, 0, Qt::TextDontClip, Name, &text_br);
+        p->drawText(tx, ty, 1, 1, Qt::TextDontClip, Name, &text_br);
     }
 
     for (auto *prop : Props) {
--- a/qucs/components/simulation.cpp
+++ b/qucs/components/simulation.cpp
@@ -119,7 +119,7 @@ void SimulationComponent::drawSymbol(QPainter *p)
     p->setPen(pen());
     p->setFont(label_font);
     QRect label_bounds;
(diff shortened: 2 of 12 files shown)


ivandi69:
So far the default zero-width curves draw fine. Shape paintings work too. Text painting doesn't draw any text if not patched. For every element lines and shapes draw fine, text doesn't if the bounding rectangle is actually a point.

dsm:
this is a qt bug https://bugreports.qt.io/browse/QTBUG-129914

and this issue fixed by this commit. https://codereview.qt-project.org/c/qt/qtbase/+/597019

We shouldn't switch Qt 6.8.0 and we should wait Qt 6.8.1 and now Qt 6.7.2 is stable enough. Qt 6.7.3 also problematic.

ivandi69:
Qt version in Arch now is 6.8.0 and in Tumbleweed 6.7.3. It might be better to fix Qucs-s and avoid that corner case.

ra3xdh:
This solution should be merged, because the Arch users are already sending bugreports. The Arch users are a big percent of Linux users. And there is no way to downgrade Qt on Arch. See #1002 

This PR is marked as WIP. @ivandi69 , are you planning to add more commits to this?

ivandi69:
@ra3xdh If all drawText calls are patched we are good to go.

ra3xdh:
You forgot `diagram.cpp` line 134 Other occurrence of `drawText` are fixed.

ivandi69:
[diagram.cpp:134](https://github.com/ra3xdh/qucs_s/blob/320e58df0a063484ad59c0d721d1e273f4839ed1/qucs/diagrams/diagram.cpp#L134) is not the same function call, it is not affected. The bug report that @dsm found explains it.

ra3xdh:
Then I am removing the WIP label and merging.

PR #1003, opened 2024-10-19
refactor: use QString() as empty QString constructor

Integral-Tech:
https://doc.qt.io/qt-6/qstring.html#QString


ra3xdh:
Thanks! Merged.

PR #1009, opened 2024-10-20
S-parameter Viewer: More fixes

andresmmera:
After more extensive testing, the following issues were spotted:

- The tool crashed if a session file was loaded when the tool had traces on the display.
- Load session feature: It had problems to keep the axes settings if the lock setting was enabled.
- Typo: Put the frequency units between parentheses.
- The limit widgets include a QFrame as a separator in order to clearly distinguish the widgets associated with a certain limit. This separator was not being removed when a limit was deleted.
- There was a problem with the marker/limit/trace lists: When the user removed a marker, the other widgets were not rearranged so the next time the user added a new marker/limit/trace, the new widgets overlapped the old ones.

All these problems have been fixed in this PR.

In addition, a global offset limit has been added. This allows the user to adjust all the limit lines in a row, and is very useful when you need to adjust a mask with a number of limit lines. As far as I know, (at least) the Keysight network analyzers have this feature.

### DIFF
--- a/qucs-s-spar-viewer/qucs-s-spar-viewer.cpp
+++ b/qucs-s-spar-viewer/qucs-s-spar-viewer.cpp
@@ -3072,6 +3072,10 @@ void Qucs_S_SPAR_Viewer::loadSession(QString session_file)
   lock_axis = lock_axis_setting;
   lock_unlock_axis_settings(false);//false means "don't change the state" inside the function
 
+  // Update chart frequency limits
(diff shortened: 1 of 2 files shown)


andresmmera:
@ra3xdh Please wait until tomorrow evening to merge this. I would like to try the Windows artifact.

ra3xdh:
Is this PR ready for merge now?

andresmmera:
Yes, it's ready.
If I find any more bugs, I will open another PR.

ra3xdh:
Merged.

PR #1028, opened 2024-10-26
Save the position / size of ExternSimDialog on exit.

### LINKED ISSUE #1012: Simulation console screen position
I realize this is a trivial Issue but can the Simulation console screen position be adjustable so I can see the plots without having to close or move the pop-up? Too many times I miss a subtle change in a plot/data because the plot is obscured. I know there is no optimum position. Another option is to have the pop-up size "minimized" unless a button is clicked to restore it's original size.

iwbnwif:
This is a simple change to store the last used position and size of the "Simulate with external simulator" dialog (class ExternSimDialog) when the user presses the "Exit" button.

This change should close #1012 and work on all platforms but I have only tested it on Linux / Qt6.

### DIFF
--- a/qucs/extsimkernels/externsimdialog.cpp
+++ b/qucs/extsimkernels/externsimdialog.cpp
@@ -31,6 +31,12 @@ ExternSimDialog::ExternSimDialog(Schematic *sch, bool netlist_mode) :
     wasSimulated = false;
     hasError = false;
 
+    QSettings settings("qucs", "qucs_s");
+    restoreGeometry(settings.value("ExternSimDialog/geometry").toByteArray());
+
+    setWindowTitle(tr("Simulate with external simulator"));
+    setMinimumWidth(500);
+
     workdir = QucsSettings.S4Qworkdir;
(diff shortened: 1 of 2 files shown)


iwbnwif:
@tomhajjar, @ainst please let me know if this addresses the issue.

ra3xdh:
I have tested this PR and noticed the following problem. When using Wayland+Qt6.7.2 the window size is restored, but the windows position is not restored. The dialog always positioned in the display center. I am not sure if this is patch issue or Wayland issue. Wayland may not allow to set the initial position of the dialog windows. I didn't test Windows and X11.

iwbnwif:
Unfortunately, I cannot get Wayland to run well on my computer. 

If you are using Gnome and have Gnome Tweaks installed, please can I ask you to check the setting Windows->Centre New Windows. The behavior of this setting appears to be different between X11 and Wayland.

ra3xdh:
I am using KDE. The KDE has a similar setting but it seems it has no effect. Let's consider this as some KDE+Wayland bug or feature. I am merging this.

Issue #984, opened 2024-10-05
Rendering performance regression since version 24.3.0


SmallJoker:
Thanks to the recent hotkey fix, I eventually updated Quc-S and noticed that scrolling the circuit area suffers from a very noticeable performance drop.

This issue appears to be caused by more inefficient rendering of many data points. In my case, when working with the sample circuit [1], any interaction *inside* the circuit area is now delayed by about 200 - 300 ms (during which an entire CPU core is 100% busy). There used to be no such delay in older versions.

**Reproduced on:**
  * Ubuntu 22.04, Polaris series GPU (amdgpu), X11

**Last known good:** 24.2.1
  * On Linux AppImage: the main process is labelled "Qt Client Leader Window". No AppRun process seen.
  * On Linux + Win64 (Wine 9.17): the main process is labelled "qucs-s 24.2.1"

**First known bad:** 24.3.0
  * On Linux AppImage: the main process is labelled  "AppRun.wrapped". No Qt process seen.
  * On Linux + Win64 (Wine 9.17): the main process is labelled "qucs-s 24.3-881d2a4"          

**Reproduction steps:**
 1. Open the sample [1]
 2. Scroll up and down in the circuit area (mouse wheel or scrollbar, or resize the window)
 3. Monitor the CPU usage

I hope you can reproduce this as well. Otherwise I'll try to profile the provided binaries *somehow*.

[1] [sample-circuit.tar.gz](https://github.com/user-attachments/files/17267316/sample-circuit.tar.gz)

SmallJoker:
Extracted the 24.3.0 (first known bad) AppImage and ran

1. `valgrind --tool=callgrind ./qucs-s`
2. `gprof2dot -s -n 5 -f callgrind callgrind.out.PROCID | dot -Tsvg -o output.svg` (<5% is ignored)

The performance analysis can be seen here:
![output](https://github.com/user-attachments/assets/139de9f9-b7a8-422b-9fbb-6428382de6de)

ra3xdh:
I confirm the problem. I have some freezing on my machine but not critical. My configuration is Arch(Manjaro)+Wayland+IntelUHD This is surely not related to Qt6 migration. It seems it has appeared after #723 where plotting engine was redesigned. @wawuwo Could you have a look? 

[...]

PR #1042, opened 2024-11-03
Improve graph rendering performance

wawuwo:
Hi! 

A quick and dirty attempt to deal with bad rendering performance of graphs.
It's not the best, but it should do the thing. I can see the actual difference when opening the sample circuit from #984 on my machine.

### DIFF
--- a/qucs/diagrams/graph.cpp
+++ b/qucs/diagrams/graph.cpp
@@ -23,6 +23,7 @@
 
 #include <QPainter>
 #include <QDebug>
+#include <QPainterPath>
 
 class Diagram;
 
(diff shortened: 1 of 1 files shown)


SmallJoker:
Thank you for the PR. I tested the [AppImage](https://github.com/ra3xdh/qucs_s/actions/runs/11654355904) a little bit, and so far it does indeed fix the performance issue. No visual issues found.

zergud:
good solution for lines!
remains drawArrowSymbols drawStarSymbols (!!) drawCircleSymbols

iwbnwif:
I am not sure if it is related to this optimisation, but I have a line being drawn to infinity if I zoom in on the chart.

image

[Test1.sch.txt](https://github.com/user-attachments/files/17617489/Test1.sch.txt)

ra3xdh:
@wawuwo , Yes were considering a similar solution with points number decimation. The freezes are gone, but some problem still exist.

* The dashed curves look odd. See the screenshots from old (before #723) and new version. This problem is not related to this fix and may be fixed by a separate PR. 

old:
image

new:
image

* I confirm the problem reported by @iwbnwif Something strange happens when using *Set diagram limits*

@zergud The circles, arrows, and stars count usually should be equal to point number. The typical usecase is a Fourier simulation (arrows) and matching of the few points of experimental data (stars). 

image

ra3xdh:
The problems with dashed curve and *Set limits* are fixed. I will merge this once the CI finishes.

ra3xdh:
Merged.

zergud:
and yet, the behavior when using stars and circles is not correct when there are a large number of points in the dataset

black bold line - circle
red - stars 

image

I think we can find a solution similar to the proposed algorithm for decimation of points
of course in another PR :)

ra3xdh:
Octave package has the same behavior as Qucs-S when using stars line style. Here is the screenshot from Octave. Blue curve is stars 2000 points, redcurve is stars 20 points. So I would keep the stars/circles in Qucs-S as is. 

image

image

PR #1043, opened 2024-11-04
refactor: replace non-empty QString constructors with QStringLiteral()

Integral-Tech:
- Qt provides a macro `QStringLiteral()`, which makes constructing QString objects from string literals more efficient.


Integral-Tech:
@ra3xdh Would you like to approve the PR? The change scope is a bit large

ra3xdh:
Merged.

PR #1048, opened 2024-11-07
Update Qt to v6.8.0 LTS

dsm:
Hi ,

MSYS2 Update Qt to v6.8.0 LTS and this PR update other OS release to Qt 6.8.0. 

In MacOS Qt 6.8.0 support minimum MacOS 12 so we drop MacOS 11 support for universal binary.


ra3xdh:
Merged.

PR #1049, opened 2024-11-08
Prepare for CDL netlist export

ThomasZecha:
Basic introductory refactoring work for code parts affected/used by the new feature CDL netlist export (https://github.com/ra3xdh/qucs_s/discussions/876)
based on https://google.github.io/styleguide/cppguide.html:
 -Prefix class attributes with a_ (much more better readability!)
 -Class member initialization via constructor member initialization list
 -No public class attributes

PR #1054, opened 2024-11-10
New Component Properties dialog

iwbnwif:
The main aim of this change is to switch from editing equations by selecting them in a table and then typing in a separate field, to entering them using free text. At the moment the editing is very free(!) and it might be that it needs to be more constrained.

The main problem I started off trying to solve is when you accidentally overwrite an existing equation by clicking 'apply' instead of 'add'. I think that free text is an easier way of entering equations, especially longer lines that didn't fit completely in the edit box.

Some basic syntax highlighting is included for Ngspice keywords only. It would also be nice to detect some basic mistakes and underline them - but this is a future enhancement.

Any rows that do not include an equals '=' are ignored and will be lost. Maybe in future it would be nice to include comments?

This is quite a substantial change and although there are a few jobs still to do, I would welcome any feedback at this stage. I can only practically test the changes on Linux at the moment so bug reports from other platforms are welcome.

TODO:
- [x]   Auto update sweep step / sweep points for log sweeps
- [x]   Translated text?
- [x]   Add special property names - i.e., for log sweeps (per decade instead of step)
- [x]   Update components from SPICE file.
- [x]   Implement highlighting.
- [x]   Have "Export" as a check box, or option list for equations.
- [x]   .INCLUDE components have multiple files
- [x]  Check for memory leaks (will be done as the final step).

A second change is to edit component properties directly in the table rather than in a separate dialog. Again, I am sure there are some corner cases that I have missed.

I have constrained my changes to just two files. The rest of the simulation and drawing is unchanged.


ra3xdh:
The issues mentioned in #1052 seems to be fixed. I have marked this PR as WIP.

iwbnwif:
> The issues mentioned in #1052 seems to be fixed. I have marked this PR as WIP. 

Thank you, I couldn't work out how to add a WIP tag, is it to just set the PR as draft?

iwbnwif:
With regard to the following comment in #1052 : 

> There exists B-type sources and EDD devices that also may have complex equations inside.

I am not sure how to detect which of the parameters has an equation as its value. There exists an <isEquation> flag for components but not for component parameters AFAICT. I have mocked this for now by allowing any parameter with a name that begins with 'I' to be edited as an equation, but this clearly isn't the answer!

ra3xdh:
> So, I propose to look at the Nutmeg script dialog as a separate change once the properties dialog has been finalised.

Yes, let's make this in another PR. I will create an issue with the list of remaining tasks.

ra3xdh:
I have just prepared the v24.4.1 release containing some important bugfxes. Now we can move forward with new features. I am intending to merge this PR in the next week.

iwbnwif:
Yes, that should be fine. I am using this branch for my normal use to see if I come across any stoppers / corner cases.

iwbnwif:
> Please resolve conflicts after merging https://github.com/ra3xdh/qucs_s/pull/1049 and rebase is necessary.

I have resolved the conflicts but I am not comfortable enough with git rebase to attempt that. If anyone can suggest the correct command line, I will give it a go.

ra3xdh:
I have found a bug in this PR. The editing the cell and pressing Enter key closes the dialog, but doesn't apply changes. The pressing OK/Apply button works as expected. 

[Screencast_20241116_122001.webm](https://github.com/user-attachments/assets/cb1cc105-1879-40c0-a39a-25a0dd0d4274)

ra3xdh:
Another issue. The sweep simulation should not be attached to ALL simulations. The drop down list should contain only real simulations, but not ALL keyword.

image

iwbnwif:
> I have found a bug in this PR. The editing the cell and pressing Enter key closes the dialog, but doesn't apply changes. The pressing OK/Apply button works as expected.

Unfortunately I am unable to reproduce this behavior. 

I think there must be OS / window manager differences and in some systems  the `QTableWidget` or `QItemWidget` keypresses are being propagated up to the dialog but I am not sure.

I have added some debug statements and perhaps you could check to see what message is shown when you press Enter in a table cell.

dsm:
> > Please resolve conflicts after merging #1049 and rebase is necessary.
> 
> I have resolved the conflicts but I am not comfortable enough with git rebase to attempt that. If anyone can suggest the correct command line, I will give it a go.

I am using gitkraken it simplifies renaming commits and squash or rebase so easy to use.

iwbnwif:
I should have printed the code as hex, but 16777220 converts to 0x1000004 which is defined as Qt::Key_Return.

We could replicate the existing dialog behavior and move to the next table row (parameter) if there is one, or apply and close if not.

ra3xdh:
Still have no effect with the latest commit. The `hasFocus()` returns `false` and `currentRow()` returns `-1` The `if` block in the keyPressEvent handler is not executed.

ra3xdh:
@iwbnwif , If nothing else helps, please integrate the following patch. It blocks the dialog accept when Return key is pressed. It seems to be a Wayland bug, because the dialog closes before the table contents is updated.

~~~
diff --git a/qucs/components/componentdialog.cpp b/qucs/components/componentdialog.cpp
index 51540806..0888262b 100644
--- a/qucs/components/componentdialog.cpp
+++ b/qucs/components/componentdialog.cpp
@@ -517,17 +517,11 @@ ComponentDialog::~ComponentDialog()
 // property table.
 void ComponentDialog::keyPressEvent(QKeyEvent* e)
 {
-  qDebug() << "Dialog key event" << e->key();
-
-  if (e->key() == Qt::Key_Return && propertyTable->hasFocus()) {
-    int row = propertyTable->currentRow();
-    if (row < propertyTable->rowCount())
-      propertyTable->setCurrentCell(row + 1, 1);
-    else
-      slotOKButton();
-  }
-  else
+  // Workaround for Qt+Wayland bug. Otherwise the Key_Retirn invokes slotOK()
+  // and table contnents not updated. Not possible to enter value.
+  if (e->key() != Qt::Key_Return) {
     QDialog::keyPressEvent(e);
+  }    
 }
 
 // -------------------------------------------------------------------------

~~~

iwbnwif:
Thank you for the patch, I was thinking along those lines too. However, please can you test https://github.com/ra3xdh/qucs_s/pull/1054/commits/14f603b52e0be0b5b5a4d5691162e6b8840b6ca0 as I would like to know if closing the persistent editor saves the changes.

If this works, I will clean up the code otherwise I will apply your patch.

ra3xdh:
The https://github.com/ra3xdh/qucs_s/commit/14f603b52e0be0b5b5a4d5691162e6b8840b6ca0 still have no effect. The Wayland issue may be more complex. I think we have to block the *Return* key processing as proposed before.

dsm:
I tested latest build on windows pressing enter don't change value too. But pressing other cell and then press value and value changed after enter or pressing tab same. So issue shouldn't be a not wayland specific. 

[value.webm](https://github.com/user-attachments/assets/aad18f91-73f9-47b8-956d-61b76ef1a979)

ra3xdh:
@dsm, Yes the KDE+Wayland has the same behavior as your screencast. I didn't check Windows.

dsm:
MacOS version don't affect this bug because pressing enter don't close panel, Btw Show section has editable text area if double pressing right after the check box icon.

dsm:
Also adding text right after check box and press enter don't close property panel.

[value.webm](https://github.com/user-attachments/assets/0319ca1b-e98c-4bd5-b344-bad594af49ed)

dsm:
`Warning: QAbstractItemView::commitData called with an editor that does not belong to this view (:0, )`

I tested Linux AppImage in WSL and I got this warning only pressing enter, pressing OK not produce this warning.

ra3xdh:
@dsm , The true Linux version also prints the same warning message after pressing Enter in the table cell.

iwbnwif:
> `Warning: QAbstractItemView::commitData called with an editor that does not belong to this view (:0, )`
> I tested Linux AppImage in WSL and I got this warning only pressing enter, pressing OK not produce this warning.

Please let me know if this message is still shown after https://github.com/ra3xdh/qucs_s/pull/1054/commits/174d2068573339f4a6b78d255d46d993aeb0b9be

I have implemented Qt::Return skipping as suggested by @ra3xdh and removed closing the QTableWidget persistent cell editor when pressing OK.

ra3xdh:
> Please let me know if this message is still shown after https://github.com/ra3xdh/qucs_s/commit/174d2068573339f4a6b78d255d46d993aeb0b9be

The warning message has gone.

dsm:
I think, first select of table select return different state after tab select, issue is not related to pressing enter.

iwbnwif:
> I think, first select of table select return different state after tab select, issue is not related to pressing enter.

Yes, I agree. I cannot reproduce the behavior that you are seeing but I do see something different on the first press, or first press after receiving focus.

dsm:
@ra3xdh KDE+X11 has the same issue in CachyOS thus issue is not wayland specific linux and windows affected only macos not affected because pressing enter didn't close dialog.

ra3xdh:
Let's accept the solution for Enter key from the latest commits. @iwbnwif , Please cleanup the debug printing from the latest commits. 

I have found an another bug with this PR. When using *Fill from SPICE* the values are not filled in the device and in the table. It should be visible in the table after user clicks OK in the dialog window.  Here is the screencast. 

Here is the source of the 1N4007 model. It worked fine with the old version:
~~~
*SRC=1N4007;DI_1N4007;Diodes;Si;  1.00kV  1.00A  3.00us   Diodes, Inc. diode
.MODEL DI_1N4007 D  ( IS=76.9p RS=42.0m BV=1.00k IBV=5.00u
+ CJO=26.5p  M=0.333 N=1.45 TT=4.32u )
~~~

[Screencast_20241118_091413.webm](https://github.com/user-attachments/assets/e8a5388a-837b-482b-976c-f1afc9282b13)

ra3xdh:
After adding `updatePropertyTabel()` at the `componentdialog.cpp` line 982 the *Fill from SPICE* works as expected.

iwbnwif:
> After adding `updatePropertyTabel()` at the `componentdialog.cpp` line 982 the _Fill from SPICE_ works as expected.

Oops! I hesitated about this because it means that even if the user presses 'Cancel' on the Component Dialog then the schematic component will still be updated. I was considering to pass a temporary component to `fillFromSpiceDialog` and then copy its contents to  `updatePropertyTable()`.

This would be more consistent with the way the rest of the dialog works but I wonder if there is a reason why the existing code works this way. This is not a feature that I have used so far so I am afraid I don't know much about it.

ra3xdh:
> it means that even if the user presses 'Cancel' on the Component Dialog then the schematic component will still be updated. 

Probably I have missed this point when designing this feature. But anyway, let's revert the operation of the *Fill from SPICE* in this PR. After fixing this problem and cleaning the debug printing I will merge this. I didn't find any other major issues.

iwbnwif:
Hopefully https://github.com/ra3xdh/qucs_s/pull/1054/commits/de5fff9999e8cc57701bd3a5304de4bff834f5db will solve the SPICE model problem. It is a little bit hacky as I should probably create copy constructors for Component and/or Property but I don't want to touch too many files.

ra3xdh:
Yes, now everything works as expected.

ra3xdh:
I have merged the PR. Stay connected, if some problems will be found by another users.

iwbnwif:
> I have merged the PR. Stay connected, if some problems will be found by another users.

Yes, I am sure there will be a lot of small problems (hopefully no big ones). The dialog deals with a lot of special cases and I tried to simplify that a bit by having kind of look up tables but I wasn't entirely successful.

I don't have a Windows development platform but I already noticed that the table line editors don't resize when you adjust the column width on Windows. This also happens on Linux but is masked by the fact that borders are not drawn around the line edit.

Also, I would like to make the compound line edit / ... button take the whole space of the table cell. At the moment there is a border that I cannot manage to get rid of.

Hopefully, these are cosmetic issues that can be corrected in time.

Issue #51, opened 2020-07-03
PoC Flatpak packaging


tinywrkb:
[This is an initial Flapak manifest for Qucs-S](https://github.com/tinywrkb/flatpaks/blob/master/com.github.ra3xdh.qucs-s/com.github.ra3xdh.qucs-s.yaml), not really tested except a basic simulation.  
It definitely needs more work, testing, and cleanup.  
This needs [Qucs base app](https://github.com/tinywrkb/flatpaks/blob/master/com.github.qucs.qucs.BaseApp/com.github.qucs.qucs.BaseApp.yaml) when building the app, so build and install it first.

I've followed the Arch Linux packaging and added every optional dependency for Qucs-S and its dependencies. So you have Ngspice, FreeHDL, GTKWave, and even also Qucs packaged.

I don't plan to publish this on Flathub myself, but I hope this might encourage you to choose Flatpak as a distribution channel.

BTW Qucs is also packaged [here](https://github.com/tinywrkb/flatpaks/blob/master/com.github.qucs.qucs/com.github.qucs.qucs.yaml).

ra3xdh:
I have recently started the generation of the Flatpak manifest for Qucs-S. I didn't find a way to launch an external application from system `/usr` from the flatpak. Therefore I decided to include the ngspice into the Flatpak package. The qucsator has dependency on ADMS that depends on Perl modules. And therefore the adding of the qucsator into the Flatpak is not a straightforward task. So I decided to restrict the Flatpak version of the Qucs-S by Ngspice support only. The manifest draft could be found here: https://github.com/ra3xdh/qucs_scripts/blob/master/FLATPAK/com.github.ra3xdh.qucs-s.yaml

tinywrkb:
> The qucsator has dependency on ADMS that depends on Perl modules. And therefore the adding of the qucsator into the Flatpak is not a straightforward task.

@ra3xdh you can use [flatpak-cpan-generator.pl](https://github.com/flatpak/flatpak-builder-tools/tree/master/cpan).

I packaged ADMS with QUCS like this
```
- name: adms
  config-opts:
   - --enable-maintainer-mode
   - PERL5LIB=/app/lib/perl5
  make-args:
   - PERL5LIB=/app/lib/perl5
  sources:
   - type: archive
     url: https://github.com/Qucs/ADMS/archive/release-2.3.7.tar.gz
[...]

PR #1056, opened 2024-11-11
Add MetaInfo

NN708:
Add AppStream MetaInfo for #51 .

### DIFF
--- a/contrib/CMakeLists.txt
+++ b/contrib/CMakeLists.txt
@@ -3,3 +3,5 @@
 INSTALL( FILES ps2sp DESTINATION bin/ )
 
 INSTALL( FILES ps2sp.1 DESTINATION share/man/man1 )
+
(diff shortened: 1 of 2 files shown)


ra3xdh:
Please move this file to `contrib` subdirectory in the repository root. The purpose of this directory is to hold different stuff mostly related to packaging. https://github.com/ra3xdh/qucs_s/tree/current/contrib

ra3xdh:
Maybe create `contrib/flatpak` subdirectory and put the both MetaInfo and manifest here.

NN708:
> Maybe create `contrib/flatpak` subdirectory and put the both MetaInfo and manifest here.

Since we will have the manifest under [Flathub](https://github.com/flathub/), I don’t think we need to duplicate it. And the MetaInfo is useful for more than just Flatpak.

ra3xdh:
Merged.

PR #1058, opened 2024-11-12
Fix build for contrib directory

NN708:
Fixed the issue that the `contrib` directory was never built.


ra3xdh:
The `ps2sp` was used as SPICE preprocessor for SPICE compatibility layer in old Qucs. It is not needed anymore and should not be installed. Also it brings perl dependency to operate. Only MetaInfo should be installed from this directory. Maybe better to move the only one remaining `INSTALL` directive to toplevel CMakeLists

PR #1063, opened 2024-11-14
Add grid color setting option

iwbnwif:
This change adds a new 'Appearance' tab to the `QucsSettingsDialog` and moves all settings related to schematic appearance to it.

A new option to change the schematic grid color is added. Currently, it requires the user to close and reopen each schematic for the change to be visible. This is because there doesn't seem to be a dynamic mechanism to update the schematic following a color change


ra3xdh:
Please resolve conflicts after merging #1049.

ra3xdh:
Merged.

PR #1064, opened 2024-11-15
Move component search box to top of component browser

iwbnwif:
This change moves the search box to the top of the component browser pane within the 'Main Dock' at the left of the application window.

The purpose is so that the UI presentation is consistent with the library browser in the same UI area and to aid feature discovery for users.


ra3xdh:
Merged.

PR #1066, opened 2024-11-17
Allow a default graph line thickness to be set

iwbnwif:
This change allows the user to set a default graph line thickness via the Application Settings dialog according to their preference.

Individual graphs can still be adjusted via the normal diagram dialog.


ra3xdh:
I was considering to introduce the similar setting. Merged.

Issue #1077, opened 2024-11-19
Wrong steps and number of points in simulation dialogs after #1054


ivandi69:
This transient simulation setup generates wrong netlist with a negative tran step: .tran -0.001 0.001 0
![image](https://github.com/user-attachments/assets/5db413ba-c27f-416b-b007-05966f1d8bf3)
![image](https://github.com/user-attachments/assets/47f4441d-3c5e-4f5e-b72c-571310af29f4)
AC linear sweep gives wrong number of points ex: 500.95:
![image](https://github.com/user-attachments/assets/20f554b5-e286-479a-8af9-5ad6ace4bdfd)
Most likely all sweeps are broken. They worked fine since #615

ra3xdh:
@iwbnwif , Please have a look. Revert an old behavior implemented in #615 if necessary.

iwbnwif:
Thanks for reporting, I will look into this.

iwbnwif:
I think this is fixed by #1079, at least for the two reported issues. As far as I can tell, the behavior for both log sweeps and lin sweeps is essentially similar to the production 24.4.1 version; however, there may be some edge cases where the user inputs invalid values for log sweeps.

Also, I notice that the list 'sweeps' do not work properly in 24.4.1 or the new component dialog. Those problems are not addressed by this fix and I will open a separate issue after more investigation.

PR #1079, opened 2024-11-19
Fix sweep steps and number of points issues  (see #1077)

iwbnwif:
This pull request attempts to correct the issues found in #1077. Primarily, these were caused by converting the numeric representation to a string using misc::str2num. This is now replaced with the built-in QString::number function.

### DIFF
--- a/qucs/components/componentdialog.cpp
+++ b/qucs/components/componentdialog.cpp
@@ -568,14 +568,14 @@ void ComponentDialog::updateSweepProperty(const QString& property)
       if (property == "Start" || property == "Stop" || property == "Points" || property == "All")
       {
         double points = str2num(sweepParamWidget["Points"]->value());
-        double step = (points - 1) / log10(fabs((stop < 1 ? 1 : stop) / (start < 1 ? 1 : start)));
+        double step = (points - 1.0) / log10(fabs((stop < 1.0 ? 1.0 : stop) / (start < 1.0 ? 1.0 : start)));
         sweepParamWidget["Step"]->setValue(misc::num2str(step));
       }
       else if (property == "Step")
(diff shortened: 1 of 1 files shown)


iwbnwif:
Please don't merge this yet as I hope to have a fix for the list sweep later today.

ra3xdh:
@ivandi69 , Please look and this if everything reported in #1077 is fixed.

ivandi69:
Installed from [here](https://github.com/ra3xdh/qucs_s/actions/runs/11920518952/artifacts/2209548623). The issues reported in #1077 are fixed.

The noise simulation dialog has "Simulation" and "Sweep Parameter" properties that are not used and not needed:
image
The DIST simulation dialog is not the same as AC simulation dialog:
image
[Test-1077.sch.txt](https://github.com/user-attachments/files/17833536/Test-1077.sch.txt)

iwbnwif:
I believe that the two problems mentioned above have been fixed in the latest commit but I have only done minimal testing as I am yet to use these two simulations.

There is still quite a bit more work to do on making list and value sweeps to work smoothly. One of the problems is that the 'Points' parameter changes name to 'Values' but retains the same position in the component->Props list. 

I think a good solution would be that the user enter list sweep points in the 'Points' line edit rather than have a separate 'Values' line edit but this will probably have some other unintended consequences that I haven't foreseen yet!

ivandi69:
DISTO and NOISE are fixed.
It would be nice to uniformise SENS DC (like parameter sweep) and SENS_AC (like ac sweep).
FOUR dialog could have a simulation selection list with only transient simulation listed.

ra3xdh:
Merged.

Issue #187, opened 2022-11-23
Crash on drag'n'drop event


ra3xdh:
I think you can reproduce this bug. I am using kubuntu 22.04 and drag the component in components tab and not drop to sch but drop again the components tab then click somewhere in the app then app crash with segmentation fault and similarly but I choose spice specific section same step but app crash with bad alloc and core dump

_Originally posted by @dsm in https://github.com/ra3xdh/qucs_s/issues/92#issuecomment-1324244749_

ra3xdh:
It seems to be a Qt version dependent issue. I don't see it on Ubuntu 18.04 using Qt5.9.5

dsm:
sometimes it gives std::bad_alloc exception sometimes it gives seg fault I'm trying to compile using qt5.15.7 

Warning: Trying to construct an instance of an invalid type, type id: 167772160
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)

ra3xdh:
I will try to debug it using Qt5.15

dsm:
I am trying to analyze valgrind but I didn't make it. Run app with valgrind make it app unusable

dsm:
![ss](https://user-images.githubusercontent.com/41967334/204030947-959bbc2a-4d68-4885-825e-b9f985844d53.png)

dsm:
any progress ?

ra3xdh:
No, I didn't make further investigations on this issue. As a workaround simple don't use drag'n'drop. Click on a device in the left panel, but don't drag it. Then contour image of the device appears on schematic following mouse pointer. And you may insert the device without using drag'n'drop.

tvrusso:
For what it's worth, I can reproduce this problem (took me a while, because I didn't understand the nature of the crash described, and don't use drag-and-drop myself).  On FreeBSD with Qt5.15.5.

Crash is very deep in Qt libraries, and I did not find anything in the stack trace that involved Qucs-s code I could look at --- it was all Qt library functions.

```
(gdb) where
#0  thr_kill () at thr_kill.S:3
#1  0x0000000801f40d44 in __raise (s=s@entry=6)
    at /usr/src/lib/libc/gen/raise.c:52
#2  0x0000000801eb9409 in abort () at /usr/src/lib/libc/stdlib/abort.c:67
#3  0x0000000801d4d259 in report_failure (err=<optimized out>, 
[...]

PR #1086, opened 2024-11-22
fix! MSVC build drag component cause a crash.

dsm:
Hi @ra3xdh , This PR fixing crash error using drag operation in MSVC build. We before semi-fixing on #187

`QVariant::save()` throw exception on MSVC but other compiler just give warning not exception. After this PR warning also gone.


ra3xdh:
Merged.

PR #1091, opened 2024-11-24
CI improvements

dsm:
Hi @ra3xdh, this PR;

- add MSVC test build to deploy.yml (can't build with qucsator_rf because autogenerated files needs post processing to replace somethings with sed tool. Adding pwsh script solve this).
- move qt5 test build to deploy.yml.
- some linker options for MSVC.


zergud:
@dsm 
i have a suggestion:
1. instead of hardcoded CI_VERSION: 24.4.99 try to take **24.4** part from current qucs-s  version
2. add git commit hash to assets name in CI build, eg **Qucs-S-24.4.99-5d47d79-macOS.dmg**
this will help distinguish CI versions

dsm:
> @dsm i have a suggestion:
> 
> 1. instead of hardcoded CI_VERSION: 24.4.99 try to take **24.4** part from current qucs-s  version
> 2. add git commit hash to assets name in CI build, eg **Qucs-S-24.4.99-5d47d79-macOS.dmg**
>    this will help distinguish CI versions

1. I retrieve the version from the VERSION file without any issues.
2. The app title includes the commit hash, but the artifact name does not. On macOS, we use Homebrew to install the snapshot builds. Since I only update the version number periodically, we always have the latest builds. However, if the commit hash is part of the file name, it would need to be updated with every new commit to the repository.

zergud:
> 1. I retrieve the version from the VERSION file without any issues.
> 2. The app title includes the commit hash, but the artifact name does not. On macOS, we use Homebrew to install the snapshot builds. Since I only update the version number periodically, we always have the latest builds. However, if the commit hash is part of the file name, it would need to be updated with every new commit to the repository.

1. Thanks a lot
2. "Since I only update the version number periodically" - version 24.4.99 never changed before next release
git hash in night builds used by many projects, for example https://downloads.kicad.org/kicad/windows/explore/nightlies

ra3xdh:
> move qt5 test build to deploy.yml

The Qt5 support will be removed in the next year. See #983. I would advise don't touch Qt5 related stuff, because it is a candidate for removing. 

>  add MSVC test build to deploy.yml (can't build with qucsator_rf because autogenerated files needs post processing to replace somethings with sed tool. Adding pwsh script solve this)

The MSVC target must present only for research purposes. The CI should only give a warning but not error if this compiler fails. I am not sure if this is possible. Many of Qucs-S contributors don't own the license for MS development tools. I also have no wish to purchase license. I remember that MSVC had some difference in C++ standards support in the past. The Qt also had some MSVC specific problems. The blocking error from CI MSVC target may potentially create difficulties in accepting contribution. And the Windows release package should be built only with free compilers (MSYS2+MinGW environment) 

> some linker options for MSVC.

This may be accepted. It doesn't interfere other compilers.

dsm:
> 2. "Since I only update the version number periodically" - version 24.4.99 never changed before next release
>    git hash in night builds used by many projects, for example https://downloads.kicad.org/kicad/windows/explore/nightlies

we have a macos brew in this link [qucs-s@nightly.rb](https://github.com/ra3xdh/homebrew-qucs-s/blob/master/Casks/qucs-s%40nightly.rb). I have to change just version number if it's increase but we have a commit hash in file name I have to change this file for every new commits. So this problem we put hash to application tile to identify which version. I can add hash to others but not macos.

dsm:
> The CI should only give a warning but not error if this compiler fails. I am not sure if this is possible.

I added a commit for this.

ra3xdh:
>  >    The CI should only give a warning but not error if this compiler fails. I am not sure if this is possible.
> I added a commit for this.

Yes, this is acceptable.

dsm:
@ra3xdh Should I add short hash to filenames (except macos because of brew)

ra3xdh:
@dsm Yes, adding hash to continuous build packages may be useful to distinguish different builds.

Kreijstal:
is there something wrong with msys2 builds?

dsm:
No, it is just an alternative build. I am experimenting.

ra3xdh:
@Kreijstal The release build must remain on the free development tools (MSYS2+MinGW). We have no plans to switch it to the MSVC. The MSVC build is only for experiments with another compilers. It is marked as `continue-on-error` and doesn't block merging.

dsm:
@ra3xdh I need to test this changes one or two more cases.

dsm:
@ra3xdh Hi, I finish my test.

ra3xdh:
Merged. The MSVC build should be available now.

PR #1112, opened 2024-12-03
AC power source amplitude fix

KrzysztofHerman:
This PR fix two issues:
1) Description string in `AC POWER SOURCE` related to the power field. It changes `Watts` to `dBm`

image

2) Calculation of the amplitude of the voltage source. In the previous version a peak-to-peak voltage was calculated and streamed to the spice netlist. According to the `ngspice` and `Xyce` specification a voltage amplitude is required. See `ngspice` manual:

image

Example from the figure above:
```
VP1 _net0 0 dc 0 ac 0.316228 SIN(0 0.316228 1MEG) portnum 1 z0 50
```


ra3xdh:
The magnitude calculation was recently fixed to match the data from Ngspice and Qucsator. See #779 and #780

KrzysztofHerman:
@ra3xdh indeed, it considers the voltage divider on the ideally matched network so the previous fix is the correct one. Shall I fix only the string in the properties window?

ra3xdh:
Yes, modify the PR.

PR #1124, opened 2024-12-05
fix path.

dsm:
Hi, I just fixed some path, AppImage first run can't find ngspice. Adding double seperator cause issue. Sometimes windows don't find but I can't find to way to delete settings (first run case)

### DIFF
--- a/qucs/main.cpp
+++ b/qucs/main.cpp
@@ -106,12 +106,12 @@ bool loadSettings()
         QFileInfo inf(QucsSettings.Qucsator);
         QucsSettings.QucsatorDir = inf.canonicalPath() + QDir::separator();
         if (QucsSettings.Qucsconv.isEmpty())
-            QucsSettings.Qucsconv = QucsSettings.QucsatorDir + QDir::separator() + "qucsconv_rf" + executableSuffix;
+            QucsSettings.Qucsconv = QStandardPaths::findExecutable("qucsconv_rf",{QucsSettings.QucsatorDir});
     } else {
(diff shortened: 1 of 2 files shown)


ra3xdh:
What happens when `QucsSettings.BinDir` has no separator on the end? Maybe to put the `QDir::separator()` between the `BinDir` and `ngspice.exe`?

dsm:
I think, I use something like that. `QDir(QucsSettings.BinDir).absoluteFilePath("ngspice.exe")`

example:
```
QString BinDir =     QucsDir.canonicalPath();
qDebug() << "Application Directory:" <<BinDir<<'\n'<< QDir(BinDir).absoluteFilePath("ngspice.exe");
```

output:
```
Application Directory: "C:/Users/MuhammetDemir/Documents/untitled/build/bin" 
 "C:/Users/MuhammetDemir/Documents/untitled/build/bin/ngspice.exe"
```

ra3xdh:
> I think, I use something like that. 

Yes, this will be also a solution.

dsm:
I try to change thanks.

dsm:
@ra3xdh I'm done with this PR.

ra3xdh:
Yes, the `QStandardPaths` is a better solution. Merging. Don't forget to test everything in a fresh VM before the release.

dsm:
Yes, I delete settings file and test all AppImage, Windows, and MacOS release but In MacOS it's hard to delete settings because OS is extreme caching and deleting not worked. I completely delete and reboot OS then everything is normal.

PR #1133, opened 2024-12-12
S-parameter Viewer: Some improvements

andresmmera:
This PR adds the following improvements and fixes to the S-parameter viewer tool:

-  **Note-taking feature**: When comparing S-parameter traces, I've found it useful to add a notepad widget so I can add my observations. The comments are saved in the .spar file and so they can check them later. Notice that I didn't write this tool, I've just simply took the note-taking Qt example [1]

- **Recent files feature**
- **Bug fix**: When loading the .spar session file, the width of the traces was 1 regardless of the previous trace setting. 
- **Data management improvement**: Now the tool saves only S-par data into the .spar files. Before this, all possible traces were saved into the .spar file, even if the user did not display them. This caused a lot of overhead, especially for large s-par data files. Now the tool only saves the raw S-parameter data.

[1] https://doc.qt.io/qt-5/qtwidgets-widgets-codeeditor-example.html

### DIFF
--- a/qucs-s-spar-viewer/CMakeLists.txt
+++ b/qucs-s-spar-viewer/CMakeLists.txt
@@ -58,9 +58,9 @@ ADD_DEFINITIONS(${QT_DEFINITIONS})
 
 #ADD_SUBDIRECTORY( bitmaps ) -> added as resources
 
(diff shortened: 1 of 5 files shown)


ra3xdh:
Merged.

PR #1136, opened 2024-12-14
Add .CSPARAM Support

Radvall:
Adds the .CSPARAM section as described in the Ngspice User’s Manual on page 66. It allows parameters from the .PARAM section to be used in the .control section. The blue color was chosen to represent this functionality. The .CSPARAM component is based on .PARAM.

_Example:_
In this circuit, the double_temp vector is created by multiplying the TEMP parameter (from .PARAM) by two. The double_temp vector is then used in a Nutmeg equation. The results of the calculation are displayed in the table and the simulation log

fig


ra3xdh:
Thanks for a new virtual device! I didn't know that Ngspice team has introduced the `.CSPARAM`.

Issue #1127, opened 2024-12-07
Simulation ran amok and created a corrupted data file crashes Qucs-S


tomhajjar:
A PLL simulation "10M_PLL_dosc_vco.sch" went amok and consumed all my PCs resources so I was forced to stop the simulation. It used up all 16GB of RAM and wrote a huge multi-GB file to spice4qucs. The resulting corrupted  "null" data file "10M_PLL_dosc_vco.dat.ngspice" crashes Qucs-S when the schematic is opened. 

Why the simulation runs amok I cannot say. I've made many changes to try and speed simulation including replacing the vco and 10M ref with simpler models.

Based on the DuSpice results showing the integrator hitting the "top" rail meaning no PLL lock so I made "10M_PLL_dosc_vco_180.sch" which flips input to phase comparator.  Under Qucs-S it ran for a very long time before Qucs-S crashed and closed. Under DuSpice the simulation ran in a reasonable time but the integrator hit the "bottom" rail again meaning no PLL lock. There are other issues in the design. The charge-pump section was just lashed together and may have issues. I haven't gotten answers to my questions.
https://sourceforge.net/p/ngspice/discussion/120973/thread/feab67cd8f/ 

I then "opened" the loop feeding the VCO with 2.5 volts. The simulation ran under DuSpice without issue but took an extremely long time under Qucs-S. Qucs-S spent more time writing it's huge data files (see attached) than DuSpice did running the simulation and plotting the results. I believe the ngspice simulation times are the same but DuSpice is an order of magnitude faster "processing" the "data" than Qucs-S.

All the netlists run under DuSpice finished in a reasonable time frame. The netlists "10M_PLL_dosc_vco_xxx.cir" "call" "VCO_dosc_5-15M.cir" and use the included .spiceinit. I made a narrow band VCO, 9-11MHz I haven't tried yet.

![2024-12-07_003817](https://github.com/user-attachments/assets/5e443973-c180-4fab-8acf-47b2789b641a)
![2024-12-07_101047](https://github.com/user-attachments/assets/709f8cd7-3f94-4ca7-a6dc-e20785d8af52)
![2024-12-07_101550](https://github.com/user-attachments/assets/402bafab-caaf-4b0d-8831-f7ef7e7b152f)
![2024-12-07_101337](https://github.com/user-attachments/assets/0311c180-4ff2-41db-ace8-65074bf45363)
[...]

PR #1139, opened 2024-12-16
Plug diagram memory leak

ivandi69:
This PR addresses a memory leak in the diagram object. With big datasets the leak can be huge.

The attached schematic can be used to reproduce the problem and test the fix. Run the simulation and monitor Qucs-s memory usage. If the diagram is deleted or the schematic tab is closed the memory is not released. If the schematic is opened again or the diagram is deleted and added again the memory consumption increases further.

[Diagram_Memory_Leak.sch.txt](https://github.com/user-attachments/files/18144284/Diagram_Memory_Leak.sch.txt)

### DIFF
--- a/qucs/diagrams/diagram.cpp
+++ b/qucs/diagrams/diagram.cpp
@@ -54,6 +54,7 @@
 #include <QDateTime>
 #include <QPainter>
 #include <QDebug>
+#include <QtAlgorithms>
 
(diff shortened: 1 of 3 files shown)


tomhajjar:
Any of this related to #1127?

I've modified the sub-circuits and schematic since and can get the circuits to simulate in a reasonable time without hogging all the resources.

ra3xdh:

PR #1154, opened 2024-12-23
Dev/tze/feature export cdl netlist

ThomasZecha:


ra3xdh:
It may be better to keep this PR open, but marking this as WIP to prevent merging by accident. This will allow to test mergeabity and apply CI. 

Considering this PR I have noticed that you have introduced `bool isCdl` parameter for `spice_netlist()` methods. This is not a good solution. It would be better to introduce an enum like the following in `spicecompat.h`:

~~~
enum SpiceDialect { SPICEDefault = 0, SPICEXyce = 1, CDL = 2 };
~~~

The `spice_netlist` should accept a single `SpiceDialect` enum value instead of chain of `bool` type values.

ThomasZecha:
Re-worked according to comments, dev-branch pushed.

PR #1180, opened 2025-01-04
Power amplifier THD vs. output voltage RMS ngspice template

ivandi69:
This PR addresses #920


ra3xdh:
Thanks for uploading an example! Merged.

PR #1186, opened 2025-01-05
Fix 'Set Diagram Limits' not updating the diagram

ivandi69:
This PR fixes the 'Set Diagram Limits' button not redrawing the diagram with the new limits.


ra3xdh:
Thanks! Merged.

PR #1201, opened 2025-01-09
Introduce optional netlisting to console

ThomasZecha:
CDL netlist will be written to console instead to file depending on qucs-s command-line parameter.

### DIFF
--- a/qucs/components/component.cpp
+++ b/qucs/components/component.cpp
@@ -840,7 +840,7 @@ QStringList Component::getExtraVariables() {
     return QStringList();
 }
 
-QString Component::getProbeVariable(bool) {
+QString Component::getProbeVariable(spicecompat::SpiceDialect) {
     return QString();
 }
 
--- a/qucs/components/component.cpp
+++ b/qucs/components/component.cpp
@@ -828,7 +828,7 @@ QString Component::getVerilogACode() {
     }
 }
 
-QString Component::getExpression(bool, bool) {
+QString Component::getExpression(spicecompat::SpiceDialect) {
     return QString();
 }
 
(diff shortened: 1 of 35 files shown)


ThomasZecha:
Can be extended to the other SPICE dialects too.

ra3xdh:
Also what happens if one generate CDL netlist and use `--run` CLI option?

ThomasZecha:
> Also what happens if one generate CDL netlist and use `--run` CLI option?

Ok, that's missing. I will add it.

ThomasZecha:
I would add something like _doCdlNetList()_ if _!run_flag && cdl_flag_, either to console or to file depending on _netlist2Console_.
The combination _cdl_flag && run_flag_ makes no sence since CDL does not contain any simulation stuff.

@ra3xdh: What do you think?

ThomasZecha:
CLI switch --cdl with the corresponding logic added

ra3xdh:
I have recently noticed that you have a leftover bool chains in `component.h`:

~~~
virtual QString getExpression(bool isXyce = false, bool isCdl = false);
virtual QString getProbeVariable(bool isXyce = false);
~~~

This should be replaced by `SpiceDialect` too. You may include the fix in this PR, or open a separate PR.

ThomasZecha:
> I have recently noticed that you have a leftover bool chains in component.h:
> 
> virtual QString getExpression(bool isXyce = false, bool isCdl = false);
> virtual QString getProbeVariable(bool isXyce = false);
> 
> This should be replaced by SpiceDialect too. You may include the fix in this PR, or open a separate PR.

Fixed.

ra3xdh:
Merged.

Issue #1228, opened 2025-02-05
Sintax error reading a 4 port touchstone file.


wsteffe:
In the annexed project, the simulation of circuit monoliticOnLTCC.sch aborts with a parsing error at line 4 of file monoliticoOnLTCC_net.s4p.
May you please look at what is wrong. The same file is imported correctly in AWR Mirowave office.

An other minor problem is that after this error pressing "Go display Page" has no effect.

Thanks
Walter

[monoliticOnLTCC_prj.zip](https://github.com/user-attachments/files/18674577/monoliticOnLTCC_prj.zip)

ra3xdh:
I guess you are using the latest version from Git. This was a bug introduced after the recent component dialog refactoring. It is not related tos2p file,, the line 4 is in the netlist.  The saving of Qucs equations became broken. I have added a fix in #1229. Wait until it will be merged and update your build. You have to fix the broken schematics by opening legacy equation properties dialog, checking the *Put result in dataset box*, and saving schematic. Or remove the equation and rewrite it from scratch if the previous point didn't help. 

>  pressing "Go display Page" has no effect

Display pages are candidate for deprecation. These pages are kept only for backward compatibility. The placing diagram on schematic is preferred way.

PR #1229, opened 2025-02-05
Library update

ra3xdh:
This PR adds two new libraries:

* `Analog.lib` contains different analog ICs: multipliers, isolation amplifiers, HF amplifiers, etc.
* `DualGateMOSFET.lib` contains dual gate MOSFET models (BF99x, BF98x). Unfortunately  I failed to find BF961, which is often used in HF circuits design. 

image

image


ra3xdh:
This PR also contains a fix for #1228

PR #1235, opened 2025-02-09
Refactor Schematic::sizeOfSelection() and its usages

wawuwo:
Hi!

Nothing fancy comes with this PR, just a bit of refactoring to replace a recurring pattern (pseudocode):
```
for (comp : all_components) {
  if (!comp->isSelected) continue;
  ...
} 
```

Member function `Schematic::sizeOfSelection()` is renamed to `currentSelection` and is made to return not only selection bounds, but also containers with pointers to selected elements, which enables to iterate over selected elements directly:
```
for (comp : currentSelection().components) {
  ...
} 
```

### DIFF
--- a/qucs/imagewriter.cpp
+++ b/qucs/imagewriter.cpp
@@ -124,7 +124,7 @@ int ImageWriter::print(QWidget *doc)
   int h = all.height();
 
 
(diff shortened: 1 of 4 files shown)


ra3xdh:
Merged.

PR #1237, opened 2025-02-09
Replace some Q3PtrList usages with QList

wawuwo:
Hi!

This patch replaces Q3PtrList with QList wherever possible. This mostly involves API (member functions) of Schematic and Mouseactions. Where it is not possible to get rid of Q3PtrList completely I inserted a "shim" which translates from QList to Q3PtrList (it's ugly, but it allows to make another step towards #748 completion).

Functionalities which are affected by this change:
- parsing schematic files
- copy-pasting
- moving of elements
- handling of "left" and "right" key presses

I've tested a bit all the mentioned above myself and found no issues, changes are cosmetic despite of their (possibly) scary appearance.

### DIFF
--- a/qucs/mouseactions.cpp
+++ b/qucs/mouseactions.cpp
@@ -105,12 +105,11 @@ bool MouseActions::pasteElements(Schematic *Doc)
     if (!Doc->paste(&stream, &movingElements))
         return false;
 
-    Element *pe;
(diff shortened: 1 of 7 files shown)


ra3xdh:
I have tested this PR and found no problems. Merged.

PR #1242, opened 2025-02-19
[WIP] XML components

andresmmera:
Based on the needs discussed in #849 and #876, this PR addresses the transition from hardcoded C++ components to their definition in XML.

To do this, [Qucs-S parses the contents of the XML libraries](https://github.com/andresmmera/qucs_s/blob/XML_Components/qucs/qucs.cpp#L973) located in `share/qucs-s/components` (as well as from the PDK directory, defined in Application Settings->Locations) and loads all library data into a [QMap structure](https://github.com/andresmmera/qucs_s/blob/XML_Components/qucs/main.h#L192).

The [Component class](https://github.com/andresmmera/qucs_s/blob/XML_Components/qucs/components/component.h) has been modified to behave like any other component. Basically, it gets the specific component information (symbol, netlist line template, etc.) through a [struct ComponentInfo](https://github.com/andresmmera/qucs_s/blob/XML_Components/qucs/main.h#L179).

Also [muParser](https://github.com/beltoforion/muparser) was integrated in order to evaluate component properties based on expressions. Consequently, it is needed to install it before building this (`sudo apt-get install libmuparser-dev`).

This is WIP at an early stage. Only some standard components have been added, it'll take time to fill the libraries and do proper testing.


ra3xdh:
Thanks for sharing! I will evaluate this and post my feedback soon. This is task for the next relese cycle.

ra3xdh:
The first point is to add muparser dependency in the `deploy.yml` CI scripts. Otherwise the CI fails.

andresmmera:
Vadim, I forgot to mention that I've made a change to the schematic file format. Since the component data in a QMap is stored by category and component name, each instance in the schematic file is now written as "Category:Component_Name".

This is very efficient for retrieving component data. I have added some logic to ensure compatibility with older schematics, but older versions of Qucs-S will not be able to open schematics created with this version (unless a patch, such as a backwards export feature, is provided).

This notation is also very convenient, IMHO, because especially when introducing PDKs, components from different libraries may have the same name.

Do you think this is a problem?

andresmmera:
Closing temporarily

PR #1244, opened 2025-02-20
Refactor Schematic::recreateComponent() and Component::recreate()

wawuwo:
Hi!

Among dozens of calls to Component::recreate() only one (!) was actually passing non-nullptr as argument i.e. recreate() had logic required only by a single specific call, which was made from Schematic::recreateComponent(). Naturally, the best option was to to move that logic out from recreate() to recreateComponent() and turn Component::recreate() into a nice zero argument function.

Also, refactor recreateComponent() to get rid of calls to malloc() and free().

### DIFF
--- a/qucs/components/DLS_1ton.cpp
+++ b/qucs/components/DLS_1ton.cpp
@@ -31,7 +31,7 @@ Component * DLS_1ton::newOne()
 {
   DLS_1ton * p = new DLS_1ton();
   p->Props.front()->Value = Props.front()->Value; 
-  p->recreate(0); 
+  p->recreate();
   return p;
 }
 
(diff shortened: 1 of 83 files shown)


wawuwo:
This one conflicts with https://github.com/ra3xdh/qucs_s/pull/1266/commits/888dfe84b39234d64e52782b003b9fe471a916c0 from #1266

ra3xdh:
What is preferred order of merge this and #1266 ?

wawuwo:
@ra3xdh, I think #1266 should go first and then I'll adapt this PR

ra3xdh:
Conflicts must be resolved after merged #1266

Issue #1245, opened 2025-02-21
Crash when delete two shorted devices


ra3xdh:
The issue has been reported in the Telegram group. Place two device on schematic and draw a wire shorting all devices. The delete everything. The application crashes. Here is a video. This may appear after one of wire refactoring PRs. @wawuwo , Please look at this. 

https://github.com/user-attachments/assets/d0908d26-1f60-44fc-8f5e-333c2c8a8ed9

ra3xdh:
Fixed by #1250

PR #1250, opened 2025-02-21
Fix crash when deleting two shorted devices

wawuwo:
Hi! This one addresses crash when deleting two shorted devices described in #1245


ra3xdh:
Thanks for the fix! The crash doesn't appear anymore.

PR #1232, opened 2025-02-06
Add two new wire forms

wawuwo:
Hi! This PR enhances wire laying by adding two new ways to create (or "routes to lay") a wire in addition to two that already present.

These two are available right now:
```
    From o---+
             |
             o To

    From o
         |
         +---o To
```
These two become available with this PR:
```
    From o---+
             |
             +---o To

      o From
      |
      +---+
          |
          o To
```

The way the wire is laid out is changed with right click while drawing a wire

PR #1253, opened 2025-02-23
Make future wire more thicker when laying a new wire

wawuwo:
Hi!

In #1232 I made ephemeral wires (i.e. "planned" or "future" wires) too thin, they are not visible behind the aim cross. This PR fixes it making future thicker and distinctly visible. Video explains it better

https://github.com/user-attachments/assets/931a5102-894a-4e02-9881-555d29f4fdfd


ra3xdh:
Thanks! Merged.

Issue #1260, opened 2025-02-27
Options: Unable to add/modify .OPTIONS parameters


tskaar:
__Description__
Currently there's no way to modify the "Name" field of the .OPTIONS blocks, causing the defaults (XyceOptionsPackage & GMIN) being the only parameters available.

Previously (v24.4.1) one was able to add/modify multiple parameters in the .OPTIONS block.

v25.1.0:
![Image](https://github.com/user-attachments/assets/dd288d27-a792-4ff1-98cb-39d7493ed511)

v24.4.1:
![Image](https://github.com/user-attachments/assets/23d2f86a-33d0-4560-9a7a-82c6b7f87ad9)
Temp added just as an example of "typical" use case for the .OPTIONS block

__Expectation__
Being able to Add/Modify the parameters per .OPTION block

__Information__
Version: v25.1.0
OS: Archlinux
Package: [AUR](https://aur.archlinux.org/packages/qucs-s)

ra3xdh:
The most probably this was a result of #1054. We missed this point while testing. @iwbnwif Could you have a look?

ra3xdh:
@tskaar You may use INCLUDE SCRIPT device as the workaround. It allows to specify the .OPTION directive. 

![Image](https://github.com/user-attachments/assets/590fac7f-438e-4002-aa6f-98fe1719a183)

tskaar:
@ra3xdh;
That's a good tip/workaround, thanks!

iwbnwif:
I think that the practical way of solving this is to treat the .OPTIONS component in the same way as we do equations. I will look at this later today.

ra3xdh:
Yes, the both .OPTIONS and .FUNC need to switch to free editing mode.

PR #1262, opened 2025-02-28
Change .OPTIONS and .FUNC components to allow free editing of contents.

iwbnwif:
Potential fix for #1260.

### DIFF
--- a/qucs/components/componentdialog.cpp
+++ b/qucs/components/componentdialog.cpp
@@ -328,7 +328,7 @@ ComponentDialog::ComponentDialog(Component* schematicComponent, Schematic* schem
   excludeList<<"Diode"<<"MCROSS"<<"MTEE"
               <<"R"<<"C"<<"Buf"<<"Inv";
 
-  // qDebug() << component->Model;
+  // qDebug() << "Component Model text is: " << component->Model;
 
   restoreGeometry(_settings::Get().item<QByteArray>("ComponentDialog/geometry"));
   setWindowTitle(tr("Edit Component Properties") + " - " + component->Description.toUpper());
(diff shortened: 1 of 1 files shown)


iwbnwif:
I have assumed that these dialogs should not have the "Simulation" and "Put result in dataset" options.

ra3xdh:
> I have assumed that these dialogs should not have the "Simulation" and "Put result in dataset" options.

Yes, the both components should not have these properties.

The .OPTIONS has another problem. The first property of this component `XyceOptionPackage` should be excluded from the free editing. It is unclear how to process this device with the new dialog. The solution may be to put a QLineEdit holding the first property. The disadvantage of this solution is that it serves only for .OPTIONS device.

iwbnwif:
Please can you point me any docs on the function of `XyceOptionPackage` and how it affects the simukation or other options.

One solution might be to create a separate component that holds this value pair. This has the advantage of only being visible in the pallette (and greyed in the schematic) when Xyce is active as the simulation engine.

ra3xdh:
The usage of .OPTIONS could be found in Xyce Reference guide, chapter 2.1.25 https://xyce.sandia.gov/files/xyce/Xyce_Reference_Guide_7.6.pdf 

I am considering to deprecate the both .OPTIONS and .FUNC components as redundant. After introduction of free editing the purpose of these devices is duplicated by include script.

iwbnwif:
Thank you for the link.

Sorry, I am missing the problem here. If I enter the following in the Options dialog, a valid netlist appears to be generated. Unfortunately, my Xyce installation is broken at the moment, so I can't test if it actually simulates.

image

Netlist snip:
```
.PARAM dBS21={dB(S[2,1])}
.PARAM dBS11={dB(S[1,1])}
.OPTIONS TIMEINT  ABSTOL = 1E-8 
```
So the netlist appears to be the same as the example in the Xyce reference guide section 2.1.25.

ra3xdh:
> Sorry, I am missing the problem here.

The user may modify the name (`XyceOptionPackage`) of the first property or event delete the first property by accident. The netlist will be wrong in such case. The first property should be with fixed name and always present. The solution may be to add validation and warning if the user press OK. The validation must be done only for `.OPTIONS` device.

iwbnwif:
Thank you for the explanation. It may be a few days before I can look at this again but hopefully that isn't too inconvenient as there is a workaround.

ra3xdh:
I am merging this. Let the possibility to edit the `XycePackage` property will be the known issue until a better fix appears.

Issue #1157, opened 2024-12-26
Subcircuit Symbol "jumps off Grid when Rotated


tomhajjar:
I thought this issue was fixed but maybe not for symbols inside subcircuits. I wanted to change symbol orientation inside subcircuit but was forced to do it at the "Schematic" level in symbols_blank.sch.

Open X_Zener.sch and save it as x_Zener_rotate.sch. Select entire Zener diode symbol and rotate "left" 3-times. Symbol "jumps off Grid. At this point you cannot "repair" the symbol. You have to select each individual line/painting, put it back on the grid, choose 1x1 grid and then "reassemble" the symbol and put everything back on the 10x10 grid...

Rotate Right would minimize some  of this...

![2024-12-26_154705](https://github.com/user-attachments/assets/dbca2afc-ae12-420b-b902-657330b14c00)

![2024-12-26_155136](https://github.com/user-attachments/assets/e83ab7de-0021-421d-b3ac-a257cf66bd17)
[Zener_rotate.zip](https://github.com/user-attachments/files/18255426/Zener_rotate.zip)

ra3xdh:
Fixed by #1266

PR #1266, opened 2025-03-08
New schematic machinery and removal of Q3PtrList

wawuwo:
Hi!

In short: this is a big rework of schematic engine which defines the way all mutations (moving, mirroring, etc.) are performed.

I've been working on this for several months (don't mind the commit history, I've rewritten it). Multiple turned-out-to-be-unpromising approaches, dead branches, dozens of cherry-picks and rebases, a lot of consumed brain fuel and finally it's seems to be if not completely done but at least working. Of course some code is kind of "crude", "dirty" and suboptimal, I hope to polish it eventually.

Internally it's a complete rework the greatest feature of which is no dependency on any of Q3PtrList-specific features. So it's another step towards replacing Q3PtrList with something modern.

From the users' side this PR offers several new features:

- non-orthogonal wires
https://github.com/user-attachments/assets/62d25721-6812-4405-926c-b8778874e992

- non-destructive mutations (connected elements remain connected after rotating, mirroring etc.)
https://github.com/user-attachments/assets/d9d24c8a-5e08-4db8-95a0-350509d571fc
https://github.com/user-attachments/assets/648ffc94-d698-4d4f-80e8-dfa6e9c00b3e

- selection of form of reconnecting-wires while moving elements
https://github.com/user-attachments/assets/be42e7da-b030-438e-8eb9-0227248e9bc3

It also possible that this PR fixes some bugs related to schematic mutations.

The bad part is that I had rewrite a lot of code in order to make different parts work like a single coherent system. A lot of opportunities to break something. Also, this PR migrates the build to C++20 which might be a breaking change.

Anyway, I'd like to get some feedback. I did a little testing on my own, but it cannot be reasonably compared to a real "battle-test". There can be bugs or new behaviour may seem unusual or even off-putting. This is an early-preview version, there is inevitably something to fix and adjust.

In short, these issues have to be addressed:
- Transition to C++20
- Bugs
- Behaviour adjustments
- overall correctness of schematics


ra3xdh:
Thanks! Looks impressive. I planned merging for v25.2.0 release cycle. 

I have found the following problem with this PR. The loading and editing of wire labels is broken. Please check this carefully. The labels on schematic are shifted after schematic load. 

image

ra3xdh:
I have marked this PR as WIP. I will make more test in the next weeks and post feedback.

ra3xdh:
Another problem found. The application crashes after attempt to mirror device horitontally (*Mirror about Y axis*). Only horizontally aligned devices are affected. 

[Peek 2025-03-08 21-04.webm](https://github.com/user-attachments/assets/2c201c1b-79c3-46ad-a9e9-7a330ae50aac)

wawuwo:
Also: pasting of wires crash the app due to assertion at schematic_element.cpp:2285

wawuwo:
@ra3xdh, what if I cram removal of Q3PtrList into in this PR? What do you think?

ra3xdh:
> what if I cram removal of Q3PtrList into in this PR? 

Yes, you can combine these two tasks. The refactoring of Schematic editor and removing Q3PtrList are closely connected.

ra3xdh:
The most of schematic editing problems are gone after the recent fixes. Now it is possible to draw a schematic. I will try to make some schematics and simulations from scratch and probably find something new. 

There is another question with new wire editor. What should be default wiring mode? Currently the diagonal wires are default. I am more inclined to set the orthogonal wires as the default. Also I recommend to put a hint label on the status bar while wiring mode is active. The hint should show a message like: *Click RMB to change wiring mode (diagonal/orthogonal)*

wawuwo:
>What should be default wiring mode? Currently the diagonal wires are default. I am more inclined to set the orthogonal wires as the default.

That's possible; as we just endlessly cycle through the wire planning functions here https://github.com/ra3xdh/qucs_s/blob/cfa2dd04459230cc3898cf450fc96af900ef2b62/qucs/wire_planner.cpp#L77 mere rearranging them will do the trick.

I'll put them in the following order:
```
1. Two-step XY
    From o---+
             |
             o To

2. Two-step YX

    From o
         |
         +---o To

3. Three-step XY

    From o---+
             |
             +---o To

4. Three-step YX

      o From
      |
      +---+
          |
          o To

5. Straight

      o From
       \
        o To
```

OK?

Also, I am almost finished with migrating from Q3PtrList to std::list, so if it's possible for you to delay a bit the testing it would be nice to check all at once

ra3xdh:
> I'll put them in the following order:

Yes, this would be acceptable

> I am almost finished with migrating from Q3PtrList to std::list, so if it's possible for you to delay

Yes, I can wait and make tests once the removing of Q3PtrList will be finished.

wawuwo:
Q3PtrList is no longer _directly_ used, but it's still used by Q3ScrollView under the hood

wawuwo:
@ra3xdh, I've patched Q3ScrollView to free it of depending on Q3PtrList Q3PtrDict and removed them completely. qt3_compat contains only Q3ScrollView now.

wawuwo:
Found some problems with paintings (text, port symbol), will provide a fix…

ra3xdh:
> I've patched Q3ScrollView to free it of depending on Q3PtrList 

Thanks! This would finally resolve #748. The Q3ScrollView class may remain in the source tree, because it doesn't depend on Q3PtrList anymore. This class would benefit from renaming and moving the CPP/H files to some new place, because it is not a Qt3Compat part anymore.

ra3xdh:
I have found another problem with schematic editor. Attempt to move the device changes the wire form randomly. The wire may become diagonal. The *Undo* operation doesn't undo this change. If the wire has attached label, the removing label crashes the application.   

[Peek 2025-03-16 21-00.webm](https://github.com/user-attachments/assets/03f3df69-44b8-4b37-9585-d73d54ea31a1)

dsm:
QAbstractScrollArea can be used instead q3scrollview last year I tried to port QAbstractScrollArea but couldn't find where it is :/ but original qucs already switch QAbstractScrollArea.

wawuwo:
@dsm 
>QAbstractScrollArea can be used instead q3scrollview

I thought about https://doc.qt.io/qt-6/graphicsview.html, but it would be a _giant_ piece of work. And as Q3ScrollView performs fine, this work looks not quite appealing to deal with… but it's still good as a backlog item I think

wawuwo:
@ra3xdh, I've made a few adjustments to the algorithm; I think now behaviour is more "natural" and like-one-would-expect:

https://github.com/user-attachments/assets/33378a5a-c5a1-426e-8ebb-8e525c30deca

"Relaying" of wires is now allowed only when schematic is changed interactively i.e. using the mouse and the depth at which the algorithm goes to relay the wires can now be limited and in fact is limited to three segments:

https://github.com/user-attachments/assets/04a0c6f2-5850-4af2-a850-4fb0bcc43088

You can adjust depth here if you don't like it (-1 means "infinite"):

https://github.com/ra3xdh/qucs_s/blob/ec90c029f0d44030c8a53dc786b3ebe775bb170e/qucs/schematic_element.cpp#L35

Undoing is also fixed… or at least _mostly_ fixed. Check this out please and let me know what you think, whether something should adjusted further.

(I still owe fixes for paintings (text and port symbol) as part of this PR 😳)

ra3xdh:
The device moving and undo are fixed now. The schematic editing mostly works as expected. I am waiting for paintings fix. I will test the subcircuit symbols creation when it will be finished.

ra3xdh:
Some assertions in schematic editor may fail. Here is a screencast how to reproduce this.

[Peek 2025-03-26 09-20.webm](https://github.com/user-attachments/assets/3e4b9d7d-cfb1-4f25-a357-6c90f5c53438)

ra3xdh:
Copy-paste of the complex schematic crashes the application., but copy-paste of the single device works as expected. Here is the screencast.

[Peek 2025-03-27 14-00.webm](https://github.com/user-attachments/assets/ee47fa7d-9616-4545-a95f-1a1a50609cce)

ra3xdh:
Library devices are placed out of grid when selecting the device. Here is a screencast. 

[Peek 2025-03-27 20-30.webm](https://github.com/user-attachments/assets/c5aa778e-02d2-483c-83c0-249354bc4bc6)

wawuwo:
>Copy-paste of the complex schematic crashes the application
>Library devices are placed out of grid 
>device is not aligned by grid after copy-paste

Fixed

>Some assertions in schematic editor may fail
>some problems with paintings (text, port symbol)

Still not; I do keep them in mind, just can't manage to find more time :( 

Also, @ra3xdh, what do you think about "flexible" wires? Should they act like this or it is better to "grow" a wire from a node to the component in situations like this?

https://github.com/user-attachments/assets/561e11bd-08a3-4745-9f6b-1ce711064cc4

It's looks kind of fancy, but I am in doubt whether it has something more than just fanciness 🤔

ra3xdh:
The fall out-of-grid and crashes are fixed after the recent commits. Once the paintings will be finished, I will test the creation of the library devices and subcircuits and report the issues if something will be found. 

> what do you think about "flexible" wires? 

The *growth* of the straight wire would be better. The straight vertical wire in this example should remain straight.

wawuwo:
@ra3xdh, I've pushed a few commits, text and port symbol are fixed now. They also provide small fixes and enhancements here and there.


> The growth of the straight wire would be better

Try setting `allowWireReshaping` to false in `keyboardMutationParams` here and compile:
https://github.com/ra3xdh/qucs_s/blob/ec90c029f0d44030c8a53dc786b3ebe775bb170e/qucs/schematic_element.cpp#L39

If didn't make any mistakes in the algorithm, it should work like you described 🤔

wawuwo:
I've found a bug: add some elements, select them and press rotate/mirror, then press Ctrl+Z — the most recently added element disappears

ra3xdh:
> Try setting allowWireReshaping to false in keyboardMutationParams here and compile:

Yes, the setting of `allowWireReshaping` to `false` acts as expected.

ra3xdh:
The build failure for Windows target is related to CI itself. It is not related to PR. See #1280

wawuwo:
I've given it another touch, undoing is OK now I presume. I've also changed the way the paintings in symbol mode are rotated and mirrored, this should fix #1157

ra3xdh:
The schematic editor works correctly for the most of cases. But I have found some new issues while editing a complex schematics and creating subcircuit symbols. I will try to determine the exact conditions to reproduce and report in the next week. I am planning to merge this PR by the end of month.

ra3xdh:
I have identified two problems using the recent build:

* Moving of the markers with mouse is not possible. Put the marker on graph, then press the LMB on marker and try to drag. It doesn't move.
* Moving of the device with mouse deletes the wire label. Moving of the device with arrow keys works as expected.

[Peek 2025-04-04 08-49.webm](https://github.com/user-attachments/assets/bede7845-edbf-4fea-a48e-e94e5ac1e434)

ra3xdh:
Wire label may jump when moving device with keyboard.

[Peek 2025-04-04 09-03.webm](https://github.com/user-attachments/assets/ea46296e-d763-4be0-8cb1-ede89f687cdc)

wawuwo:
Vadim, I've fixed the bugs you previously mentioned, turns out I completely _forgot_ to process markers.

Labels are "shifted" towards "stable" node when wires are laid anew during movement:

https://github.com/user-attachments/assets/6182b04e-802a-49e1-b4c9-6bb991510a1e

Finding a place for label root on a new wires is possible but requires some effort because at the time of deciding that a wired connection is going to be laid anew, it's unknown how new wires will be laid out. It's just "these wires are obsolete, so to keep things intact connect this and this point with wire". The actual form of the new wired connection is determined at the moment of connecting the points.

So for now I decided to go the easy way. As I understand a label which is shifted to a node is semantically the same, just looks different. But I don't know… if it's critical, I can think something up.

ra3xdh:
I confirm that the markers and label jump are fixed using the recent commits. I have encountered a crash when deleting a fragment in the complex schematic, but I cannot catch it anymore. Probably the recent commits has fixes this problem too. I will report if I find conditions to reproduce. 

Nevertheless, the new schematic editor is almost ready. I am planning to merge this PR soon. Don't forget to put a indicator of the wiring more (orthogonal/diagonal) on the status bar at the bottom of main window.

wawuwo:
>… wiring more (orthogonal/diagonal) …

Just to clarify the things: these are not modes, but different ways to connect two points offered by wire planners. Wire planners take two points (e.g. A and B) as input and return a sequence of points which when connected one by one establish a wired connection between A and B. So it's possible to have a wire planner which offers V-shaped, П-shaped, etc. connection between two input points. Diagonal wires are just result of work of the simplest wire planner 🙂 


>to put an indicator on the status bar

I'm okay with the idea, but what's the purpose of it? I mean when a wire is being created its form is visible on the screen… What use case I don't get?

>I am planning to merge this PR soon

What about the settings like`allowWireReshaping`? Shouldn't the setting value from [here](https://github.com/ra3xdh/qucs_s/pull/1266#issuecomment-2765366672) be applied by default?

Just in case, here are descriptions:
- `allowWireReshaping` — enables or disables "flexible" wires
- `allowWireRelaying` — enables or disables "laying wires anew"
- `wireRelayingDepth` — controls how deep to go a long the wire to find "stable" node when laying wire a new. This setting has no meaning without `allowWireRelaying=true`

ra3xdh:
> but what's the purpose of it?

The idea is to give user a hint to press RMB to change the wiring. The caption may be like *Orthogonal: press RMB to switch to diagonal*

> What about the settings like allowWireReshaping?

The `allowWireReshaping` in the `keyboardMutationParams` should be set to `false`.

wawuwo:
>The idea is to give user a hint to press RMB to change the wiring. The caption may be like Orthogonal: press RMB to switch to diagonal

Got it! Somethings like this?

https://github.com/user-attachments/assets/c6a672a8-c40d-4416-b45a-79c7cd5e0a96

ra3xdh:
> Got it! Somethings like this?

Yes, this would be acceptable.

ra3xdh (approved):

ra3xdh:
I have found a bugs with export schematic to graphic files.

* Right click on diagram and select *Export to image* Then select SVG and open SVG with inkscape. The image will be shifted.
*  Select a fragment of schematic, then select in main menu *File->Export to image*, in the dialog select *Export to selection* The SVG will have a large white filed. 
* The same for export all schematic.
* PNG export of schematic is partially affected, PNG export of diagram is also shifted

Here is the test schematic: 
[tst.sch.gz](https://github.com/user-attachments/files/19645084/tst.sch.gz)

ra3xdh:
The problem with diagram export has been fixed by the recent commits.

ra3xdh:
I am merging this PR. The remaining things may be fixed in a separate patches. @wawuwo , Please stay connected. Some feedback may appear after the users will test the continuous build.

ra3xdh:
The earliest possible release date for v25.2.0 is end of July. The latest is September. Everything should be ready by the  July. We have plenty of time to implement remaining things.

PR #1275, opened 2025-03-21
Fix Xyce IC for Tline

### LINKED ISSUE #1274: TLine in Xyce schematic not accepted
This schematic follows in a abort for Xyce simulation:
Image

The generated netlist contains:
`T1 _net2 0 out 0 Z0=50 Td=0.25N F=1E9 NL=0.25 IC=0, 0, 0, 0 
`
For me it looks that the initial condition is not valid for Xyce regarding their Reference Guide:

```
Instance Form T<name> <port 1 (+) node> <port 1 (-) node>
+ <port 2 (+) node> <port 2 (-) node>
+ Z0=<value> [TD=<value>] [F=<value> [NL=<value>]]
```

ra3xdh:
This PR fixes #1274 and makes simulation of TLine with Xyce possible. Define unused parameters as empty strings. 

image

### DIFF
--- a/qucs/spicecomponents/LTL_SPICE.cpp
+++ b/qucs/spicecomponents/LTL_SPICE.cpp
@@ -121,7 +121,11 @@ QString LTL_SPICE::spice_netlist(spicecompat::SpiceDialect dialect /* = spicecom
         s += QStringLiteral(" NL=%1").arg(Nl);
     }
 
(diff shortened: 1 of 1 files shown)


dwarning:
Thank you, it helps.
But why not also skip the initial conditions in the component property box with "dialect" switch for Xyce?

dwarning:
OK, thank you.

PR #1296, opened 2025-04-09
fix: Make label root follow its host node when the node is moving

wawuwo:
I forgot to override moveCenter() in Node and it used inherited implementation which doesn't account label.

Fix ra3xdh#1293

### DIFF
--- a/qucs/node.cpp
+++ b/qucs/node.cpp
@@ -98,3 +98,12 @@ Element* Node::other_than(Element* elem) const
 
   return other == connections.end() ? nullptr : *other;
 }
+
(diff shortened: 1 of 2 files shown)


wawuwo:
@dsm can you have a look please? it seems that something is wrong with build scripts after recent update

dsm:
https://github.com/jurplel/install-qt-action/issues/281 problem is related to this. and not script but I 'll look workaround.

wawuwo:
Issue seems to be resolved by itself (another PR has just passed all checks with no problems).

PR #1306, opened 2025-04-13
Fix rectangle bounding box mismatch

wawuwo:
Here are fix for the 1305 and improvement of paintings selecting.


wawuwo:
There is also a problem with selecting of an ellipse: it cannot be selected by clicking on its segment from 0 to 90 degrees (counting clockwise from ellipses horizontal axis), click has to be made slightly closer to the center to select the ellipse. I don't have a solution yet.

PR #1307, opened 2025-04-15
XML-based component prototyp

ThomasZecha:
Prototyp for further migration of hard-coded devices to xml, covers so far:

1. Processing of xml-based components up to graphical representation in the qucs-s component panel
2. Support of shared symbols
3. Automatic xsd code generation
4. Deployment of xsd-dependencies to CI


qucs/module.cpp
```diff
@@ -31,29 +32,49 @@
...
+    a_xmlComp()
+{
 }
 
 // Destructor removes instance of module object from memory.
-Module::~Module () {
-    if (icon != nullptr) delete icon;
+Module::~Module()
+{
+    if (a_icon != nullptr) delete a_icon;
```
wawuwo, line 53: It's safe to use `delete` on null pointers 🙂

ThomasZecha: This logic hasn't been changed by this PR, only _a__ prefix for class attributes where introduced.
Much better would be avoiding raw c++ pointer by using RAII with smart-pointer, but that's another story.

wawuwo: >This logic hasn't been changed by this PR, only a_ prefix for class attributes where introduced.

😳 I missed it entirely

wawuwo:
@ra3xdh, I see that `Component` interface has methods
```
  virtual QString netlist();
  virtual QString spice_netlist(spicecompat::SpiceDialect dialect = spicecompat::SPICEDefault);
  virtual QString cdl_netlist();
  virtual QString va_code();
  virtual QString vhdlCode(int);
  virtual QString verilogCode(int);
```
which seem to be somewhat similar. Isn't it possible to replace them with a single method named like "serialize" or similar which takes a simulator (more generally — "output type") identifier and returns a corresponding string? Internally it would work as following:
1. Look for a "template string" for a given simulator identifier
2. Prepare an "environment" — a map consisting of component properties and some other data (e.g. node names, etc.)
3. Fill the template using this environment

Something like (pseudocode):
```
QString serialize(OutputTypeEnum outType)
{
    if (!hasTemplateFor(outType) throw "unsupported"
    auto template = templateFor(outType);
    auto env = environmentFor(outType);
    return template.fill(env);
}
```

It would provide a way to create and reuse generic logic. Unfortunately I lack a proper domain knowledge to assess myself how viable this idea is… what do you think?

This PR as I understand lays some foundation for such an approach — components' XML-definitions contain templates for CDL and ngspice.

ra3xdh:
The unification of netlist entry generation is not possible. Qucsator syntax was designed to have the same netlist entry template for every device. There is one common `Component::netlist` method for all devices with few exception. But every SPICE device requires an unique entry. Some of the SPICE devices are replaced by complex subcircuit and some devices need to look into the parent schematic to generate netlist. The netlist generation from common template is possible only for small subset of SPICE devices. So, I propose to keep the netlisting as is.

ThomasZecha:
@ra3xdh: Any plans for merging this PR?

ra3xdh:
> Any plans for merging this PR?

I didn't test it yet. Maybe in the next month.

ThomasZecha:
Please note that this is only a prototype as a feasibility study.

PR #1315, opened 2025-04-18
Refactor usages of Wire

wawuwo:
Here are little changes here and there to make usage of Wire a bit cleaner. Mostly by removing explicit actions like "set this node as port1, update coordinates x1, y1, update coordinates of center, etc." in favour of API calls which hide all the details. Another goal was to get rid of direct access to member fields x1, y1, x2, y2.

### DIFF
--- a/qucs/healer.cpp
+++ b/qucs/healer.cpp
@@ -125,9 +125,9 @@ QPoint GenericPort::center() const
         case PortType::Component:
             return m_comp->center() + QPoint{m_port->x, m_port->y};
         case PortType::WireOne:
-            return QPoint{m_wire->x1, m_wire->y1};
+            return m_wire->P1();
         case PortType::WireTwo:
-            return QPoint{m_wire->x2, m_wire->y2};
+            return m_wire->P2();
     }
     assert(false);
 }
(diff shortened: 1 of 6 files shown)


ra3xdh:
Thanks! Merged.

ivandi69:
With this PR qucs-s segfaults every time a wire or wired component is moved. Reverting the merge fixes the issue.

wawuwo:
>every time a wire or wired component is moved

Without any preconditions i.e. just insert a wire to empty schematic, move it and see the crash?

ra3xdh:
Open any schematic, drag the device and you will see the crash.

[Peek 2025-04-19 16-13.webm](https://github.com/user-attachments/assets/74ff46de-ed7b-41c2-93f8-5827fd66f6f4)

wawuwo:
Well, strangely enough, it doesn't happen on MacOS + clang, but does happen on Debian + GCC

wawuwo:
OK, I've found out the cause. A child's mistake — uninitialised member fields

wawuwo:
I've opened a PR with the fix. I'm sorry for the inconvenience. Mac is my primary machine, so I didn't spot the crash. 🙂

Issue #1291, opened 2025-04-08
Refactor Schematic::activateSpecifiedComponent


wawuwo:
The body of `Schematic::activateSpecifiedComponent` uses [legacy API](https://github.com/ra3xdh/qucs_s/blob/d2c8868278e517e47d536856b056dc0ed49973fa/qucs/schematic_element.cpp#L1973) to obtain bounds of a component and clumsy chain of `if`s to check whether a point lies within the bounds. It also returns on a first component which passes the if-chain, i.e. it's basically interested only in first suitable component to do the things to.

What has to be done:
- the loop has to be replaced with `ranges::find_if` and a lambda
- instead of `Bounding` the `boundingRect` method has to be employed to obtain component bounds
- call to `contains`on QRect returned by `boundingRect` should be preferred to `if`s

The _perfect_ way to present this is in two commits: one for switching to new API, another for replacing the loop with `find_if` and a lambda:
- "refactor: Replace legacy Bounding() with boundingRect()"
- "refactor: Rewrite activateSpecifiedComponent() with no loop"

PR #1322, opened 2025-04-22
Refactor Schematic::activateSpecifiedComponent()

poh0:
Made changes requested in #1291 
- Refactor Schematic::activateSpecifiedComponent() to use boundingRect() method instead of legacy Bounding() and rewrite the function to use std::ranges::find_if and a lambda instead of a for loop.

Now, I wasn't sure if I should clang-format the whole function, so I just used the old style.

Let me know if there's any changes to be made.


wawuwo:
>Now, I wasn't sure if I should clang-format the whole function, so I just used the old style.

It's OK, we've tried to install some formatting standards but it didn't actually go off, so it's good the way you've done it.

PR #1324, opened 2025-04-22
fix: Color text in editor depending on lightness of BGColor

### LINKED ISSUE #1323: Editor: Dark theme and light background causes text to be "invisible"
Hi,

When using a dark-theme (I'm using KvantaDark as a reference here), but with a light background (say default BGColor), one ends up with the text in the editor becoming invisible, e.g.:

And just as a sanity check, here it is with the text-highlighted:

With the referenced PR one gets the "correct" behavior:

tskaar:
In a scenario where one uses a dark theme, but a light background color, the default text color is then white due to the theme. This applies explicit coloring depending on the background color in settings.

fixes: #1323


ra3xdh:
Thanks for reporting issues and fix! Merged.

PR #1336, opened 2025-05-02
Fix for issues 1223 regarding CDL netlisting

ThomasZecha:
-Fixed unresolved subcircuit instance model name in CDL netlisting:
 As an optional post-processing step during CDL netlist export:

 1. Found all subcircuit instances in the current schematic.
 2. For every instance from 1. check whether the referenced subcircuit with the same number of pins exists in the current 
 schematic.
 4. For every instance from 2. where the subcircuit don’t exist do: a. Check whether the second letter of the instance name refer a CDL primitive device (R, C, L, M, Q, ...) c. If a. holds remove the first letter from the instance name for the CDL-netlisted instance name.

-Introduced command-line parameter and CDL settings dialog for resolve spice prefix parameter

-Fixed misleading function name

### DIFF
--- a/qucs/dialogs/librarydialog.cpp
+++ b/qucs/dialogs/librarydialog.cpp
@@ -509,7 +509,7 @@ void LibraryDialog::slotSave()
                                 "Check these components: %2 \n")
                     .arg(Doc->getDocName()).arg(err_lst.join("; ")));
         }
-        kern->createSubNetlsit(ts,true);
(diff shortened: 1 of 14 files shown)


ra3xdh:
The checkable menu is not the best place to hold the settings. The better way is to add a special dialog related to CDL netlister settings.

ra3xdh:
Merged.

PR #1309, opened 2025-04-16
[WIP] S-parameter viewer update

andresmmera:
I've updated the S-parameter viewer and added some features;

 - **Phase Display**: Added a right y-axis to the rectangular plot for displaying phase information alongside magnitude

- **New Chart Types**:

   - Smith chart

   - Polar chart

   - Group delay

 - **New data formats handling**: Now it reads Qucs-S data files (.dat, .ngspice.dat) in addition to Touchstone files.

- **Live data monitoring**: Implemented real-time monitoring of project data files. The display automatically updates as simulation data changes, enabling interactive filter tuning with visual limit line feedback (see attached video).
- **Codebase Improvements**: Restructured the code for better readability and addressed various bugs.

Video: https://github.com/user-attachments/assets/5aecc97a-99e8-4a95-9314-f07d070a83a7

I'm marking this as WIP. It's likely that some bugs may be found soon...

PR #1339, opened 2025-05-03
S-parameter viewer: Fixes

andresmmera:
This PR addresses some issues detected after merging PR #1309:

- Fix save / load feature
- When the user removes a file, ensure that the dataset is removed.
- Fix Rectangular plot y-axis autoscale
- Fix duplicate addition when a file was created in the directory the program is watching
- Add import / export functions to the chart widgets to save/load the specific settings of each viewer in/from the .xml file
- Assign object names to the docks in order to avoid warnings when saving the tool state.


ra3xdh:
Thanks! Merged.

PR #1346, opened 2025-05-05
Project view: Open files with the system default program

andresmmera:
I've found that if a user double-clicks a file in the "Others" section of the Project View, Qucs-S loads the file as a text file.

As far as I know, Qucs-S tries to find a suitable program by inspecting QucsSettings (see: https://github.com/ra3xdh/qucs_s/blob/current/qucs/qucs.cpp#L2840). If it doesn't find a suitable application, it opens the file as a text file.

In this PR, I've added code to use QDesktopServices if the previous method fails. Additionally, Touchstone files are now opened with the Qucs-S S-parameter viewer.

**Before this PR**

https://github.com/user-attachments/assets/478a21e4-6b74-4e2c-b9d6-fbab5019abdb

**After this PR**

https://github.com/user-attachments/assets/8d4b21e2-b671-45ff-8126-e340c6081da0

### DIFF
--- a/qucs-s-spar-viewer/main.cpp
+++ b/qucs-s-spar-viewer/main.cpp
@@ -104,11 +104,22 @@ int main( int argc, char ** argv )
 
   Qucs_S_SPAR_Viewer *qucs = new Qucs_S_SPAR_Viewer();
 
-  if (argc > 1) { // File directory to watch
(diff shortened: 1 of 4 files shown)


ra3xdh:
Thanks! Merged.

PR #1356, opened 2025-05-08
Refactor property

wawuwo:


wawuwo:
@ra3xdh could please take a look? I'm interested if it goes in the correct direction so far.

I've adapted this component as an example
image

ra3xdh:
> could please take a look? I'm interested if it goes in the correct direction so far.

Yes, I will try to compile and test this patch. Need to test save/load of the components. 

> I've adapted this component as an example

The `LTL_SPICE` is a bad candidate. This device is SPICE only and the whole device is visible only if SPICE simulator is selected. This device should be kept unchanged.  The good candidate is Transient simulation. The all-simulator properties are only Type,Star, Stop,Points, MaxStep, and InitialDC. All other properties are Qucsator-only and have no effect for SPICE. These properties should be hidden if SPICE simulator is selected to prevent accidental setting by user.

ra3xdh:
Also look at the output of the git grep for Qucsator-only properties:
~~~
git grep -n "Qucsator only"                                                                                                                                       ✔  11s  
qucs/components/capq.cpp:74:                QObject::tr("simulation temperature in degree Celsius (Qucsator only)")));
qucs/components/cccs.cpp:64:  Props.append(new Property("T", "0", false, QObject::tr("delay time (Qucsator only)")));
qucs/components/ccvs.cpp:64:  Props.append(new Property("T", "0", false, QObject::tr("delay time (Qucsator only)")));
qucs/components/indq.cpp:75:        QObject::tr("simulation temperature in degree Celsius (Qucsator only)")));
qucs/components/resistor.cpp:30:    QObject::tr("simulation temperature in degree Celsius (Qucsator only)")));
qucs/components/resistor.cpp:36:    QObject::tr("temperature at which parameters were extracted (Qucsator only)")));
qucs/components/switch.cpp:36:        QObject::tr("simulation temperature in degree Celsius (Qucsator only)")));
qucs/components/switch.cpp:40:        QObject::tr("Resistance transition shape (Qucsator only)")+" [abrupt, linear, spline]"));
qucs/components/vccs.cpp:66:  Props.append(new Property("T", "0", false, QObject::tr("delay time (Qucsator only)")));
qucs/components/vcvs.cpp:69:  Props.append(new Property("T", "0", false, QObject::tr("delay time (Qucsator only)")));

~~~

ra3xdh:
The hiding properties works wrong. The display checkbox has no effect. After the simulator is switched, the properties are shifted. See the attached screencast. 

[Uploading Peek 2025-05-08 16-59.webm…]()

ra3xdh:
This componentdialog.cpp may fill properties by number of the table row if *Apply* button is clicked. If something is hidden, the properties will be shifted.  The fill by name is not possible because properties may have duplicate name like "File". I am inclined to forbid duplicate properties names and use the new `type` field instead. The aternative way also may be found.

wawuwo:
🤔 I'll look into that

wawuwo:
@ra3xdh I've updated component dialog and adapted other components you mentioned earlier. It seems the dialog works OK now.

ra3xdh:
I will provide a review in the next few days. I am planning to send you a PR implementing handling of File type properties.

ra3xdh:
> I am planning to send you a PR implementing handling of File type properties.

This may not be needed. I have to test the recent changes.

ra3xdh:
Now everything works as expected. Merging. I will provide a patch with the revision of *File* type properties later.

PR #1357, opened 2025-05-08
Fix incorrect default diode model parameters in .MODEL Section component

Radvall:
The default diode model parameters in the `.MODEL Section` component contained ngspice-incompatible parameters 'BF' and 'Vbf' . This caused warnings during simulation:

изображение

This commit replaces:
- 'bf' with 'bv' parameter (Reverse breakdown voltage)
- 'vbf' with 'n' parameter (Emission coefficient)

The new default model is:
.model DIODE1 D (bv=50 is=1e-13 n=1.5)


ra3xdh:
Thanks! Merged.

PR #1359, opened 2025-05-09
Fix BJT and MOSFET qucsator netlist generation

andresmmera:
I've noticed that the Qucs-S examples with BJT and MOSFET devices do not work in qucsator.

I'm aware that one shouldn't use qucsator for these devices... Anyway, I fixed the netlist generation to skip the "UseGlobTemp" which qucsator seems not to understand and crashes.

**Steps to reproduce**

1) Open "examples/qucsator/single_balanced.sch" (MOSFET) or "examples/qucsator/classic_osci.sch" (BJT)

Selection_021

Selection_022

### DIFF
--- a/qucs/components/bjt.cpp
+++ b/qucs/components/bjt.cpp
@@ -145,8 +145,12 @@ QString BJT::netlist()
   s += " "+Ports.at(1)->Connection->Name;  // connect substrate to collector
 
   // output all properties
-  for(const auto& p2 : Props)
+  for(const auto& p2 : Props) {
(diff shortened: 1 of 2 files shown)


ra3xdh:
You are probably using wrong Qucsator version. This is fixed in QucsatorRF by this commit: https://github.com/ra3xdh/qucsator_rf/commit/3fbc583fdafb8e1bcd61b921e74e6142919b56b8 It should show warning instead of error and continue simulation. 

image

andresmmera:
Oh, I'm sorry.
Thank you!

Issue #1372, opened 2025-05-18
Qucs-s should not default to LEVEL=2 for Xyce diode


tvrusso:
A long time ago, the Xyce level 1 diode was strictly SPICE3F5 compatible and was  missing parameters that other simulators (ngspice?) had.  The level 2 had those parameters, and so apparently Qucs-S defaults to emitting LEVEL=2 in all diode model cards for Xyce.

As it turns out, the Xyce level 2 diode was a poorly implemented "PSpice-compatible" mode that had some parameters that worked properly, some parameters that were just accepted but ignored, some parameters that were accepted and were used incorrectly, and otherwise didn't really make the model PSpice compatible at all.

Back in 2022, the Xyce level 1 diode was reworked, adding many of the features that were supposedly available in the level 2 diode, and several more (including sidewall effects, etc.), and the level 1 diode is more ngspice-compatible than it was prior to 2022.   In almost every case where the level 2 recognized and used a parameter properly, it would have fallen back to the level 1 if that parameter had been left at defaults, and so in all of those cases the code was made to apply to both level 1 and level 2.  In cases where the level 2 accepted a parameter but ignored it, the code was made to use it properly and fall back to level 1 equations when left at defaults.   

In fact, there are now only a couple of lines of code that make Xyce's level 2 diode different from the level 1, and those few lines are actually suspect in that they just change the way temperature correction to junction voltage and junction capacitance are computed, and are probably wrong  (c.f. [this issue report](https://github.com/Xyce/Xyce/issues/138)).  Using the level 1 is highly recommended now,  and the level 2 should probably be removed from Xyce (though whether the team will do that remains to be seen).   

The issue of these things being wrong was actually pointed out to us by a qucs-s user, precisely because qucs-s is using the level 2 diode by default.  While that might have been necessary in the past because of the limitations of the Xyce diode prior to 2022, it should not be the case anymore.

The code responsible for this in qucs-s is on line 163 of qucs/extsimkernels/qucs2spice.cpp, where there's this conditional:
```
    if (xyce) s += QStringLiteral(".MODEL DMOD_%1 D(LEVEL=2 %2) \n").arg(name).arg(mod_params);
    else  s += QStringLiteral(".MODEL DMOD_%1 D(%2) \n").arg(name).arg(mod_params);
```

I suggest that you remove the conditional part for Xyce and just use the line in the "else" clause.

tvrusso:
[...]

PR #1373, opened 2025-05-18
Stop forcing LEVEL=2 into Xyce diode conversions

tvrusso:
Qucs diode models use some parameters that the Xyce level 1 diode did not support prior to 2022, and the "PSpice-compatible" level 2 diode did have those parameters.

Xyce's level 1 diode has been reworked to include those parameters and more, and most of the differences between level 1 and level 2 are gone.  In fact, the only differences that remain are in the computation of junction voltage and junction capacitance, and it is believed that they are incorrect anyway.

This commit removes the code that forces LEVEL=2 into diode models converted from Qucs models to Xyce netlists.

Closes GH-1372


ra3xdh:

PR #1378, opened 2025-05-20
Open multiple schematics from the Content View tab

andresmmera:
This PR enables the ProjectView class to open multiple files in a row by holding the CTRL or SHIFT key. The context menu has also been modified to allow for the removal of multiple files. 

The 'Rename' and 'Duplicate' options are disabled when multiple files are selected.

https://github.com/user-attachments/assets/1f5d19ab-78c6-472f-9825-8de0c5153015

### DIFF
--- a/qucs/projectView.cpp
+++ b/qucs/projectView.cpp
@@ -78,16 +78,16 @@ ProjectView::refresh()
   header << tr("Content of %1").arg(m_projName) << tr("Note");
   m_model->setHorizontalHeaderLabels(header);
 
-  APPEND_ROW(m_model, tr("Datasets"), QString(""));
-  APPEND_ROW(m_model, tr("Data Displays"), QString(""));
-  APPEND_ROW(m_model, tr("Verilog"), QString(""));
-  APPEND_ROW(m_model, tr("Verilog-A"), QString(""));
-  APPEND_ROW(m_model, tr("VHDL"), QString(""));
-  APPEND_ROW(m_model, tr("Octave"), QString(""));
-  APPEND_ROW(m_model, tr("Schematics"), QString(""));
-  APPEND_ROW(m_model, tr("Symbols"), QString(""));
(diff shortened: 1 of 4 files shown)


dsm:
Hi @andresmmera, I suggest removing the APPEND_ROW macro, we can use inline function or lambda function. 


```cpp
inline void appendRow(QStandardItem* parent, const QString& data0, const QString& data1)
{
    auto* col0 = new QStandardItem(data0);
    auto* col1 = new QStandardItem(data1);

    col0->setFlags(col0->flags() & ~Qt::ItemIsSelectable);
    col1->setFlags(col1->flags() & ~Qt::ItemIsSelectable);

    QList<QStandardItem*> row{ col0, col1 };
    parent->appendRow(row);
}
```


```cpp
auto appendRow = [](QStandardItem* parent, const QString& data0, const QString& data1) {
    auto* col0 = new QStandardItem(data0);
    auto* col1 = new QStandardItem(data1);
    col0->setFlags(col0->flags() & ~Qt::ItemIsSelectable);
    col1->setFlags(col1->flags() & ~Qt::ItemIsSelectable);
    parent->appendRow({ col0, col1 });
};
```

andresmmera:
Hi @dsm

Thanks for the suggestion!

I've just converted the APPEND_ROW and APPEND_CHILD macros into inline functions locally (not pushed yet).

(Please excuse my ignorance...) As far as I could see, macros and inline functions improve the performace when a piece code is intended to be called many times. I think this is not the case in the ProjectView class, so these macros can even be regular functions, Am I right?

I also have another local commit with the APPEND_ROW and APPEND_CHILD macros as regular member functions. 

@ra3xdh Should we keep the macros or turn them into regular functions or inline functions?

ra3xdh:
Macros is a bad practice in C++ and should be avoided. The inline functions serve as a good replacement.

andresmmera:
Great!

Pushed commit with APPEND_ROW and APPEND_CHILD macros as inline functions.

Issue #1385, opened 2025-05-29
Marker jumps to start of diagram when added if diagram has a very narrow custom time/x interval


njascgil:
As the title says, if I try to add a marker to a diagram which was configured for a very narrow time interval, the marker does not stay where I click but jumps instead to the left.
Additionally, with the same narrow time interval, the diagram appears to show some artifacts on top of the left y axix.

![Image](https://github.com/user-attachments/assets/547a12da-57c4-46e8-a6a0-8e1061a5d007)

I am attaching a file/simulation where both issues I am facing should be visible. Just run the simulation with the Qucsator engine and try to add a Marker somewhere.

[marker_bug.zip](https://github.com/user-attachments/files/20500800/marker_bug.zip)

ra3xdh:
I confirm this. This may be related to the recent redesign of diagrams and schematic #1266 @wawuwo , Could you check?

The workaround is to set the X-position in the marker properties dialog:

![Image](https://github.com/user-attachments/assets/dcace85b-ca5d-491d-b3aa-a7b9f57efbb4)

wawuwo:
I’ll take a look at this on the weekend 29.05.2025, 11:37, "Vadim Kuznetsov" ***@***.***>:ra3xdh left a comment (ra3xdh/qucs_s#1385)
I confirm this. This may be related to the recent redesign of diagrams and schematic #1266 @wawuwo , Could you check?
The workaround is to set the X-position in the marker properties dialog:
image.png (view on web)

wawuwo:
I have given it a look, it seems that the problem may be somewhere in "diagraming" code and is not related to #1266. The same behaviour can be observed in rather old Qucs-S 24.1. "Diagraming" code is quite messy, I think solution will take some time.

ra3xdh:
Fixed by #1391

PR #1391, opened 2025-06-01
Fix wrong marker root position

wawuwo:
Hi! This is fix for #1385.

This was hard to catch tbh :)

In short the bug is caused by integer overflow when calculating distance between the click and datapoint. In some configurations `x` becomes so large that its square overflows giving wrong result i.e. a datapoint appears to be closer to the click than it actually is.
```
    d = x*x + y*y;
    if(d < dmin) {
```


ra3xdh:
Thanks! Merged.

PR #1397, opened 2025-06-05
graphictext: store transformed boundingRect

tskaar:
Currently, when rotating a graphictext object, the text itself gets rotated, however the boundingRect of the object doesn't get the rotation applied. Hence the "selection box" is not where you expect it to be. 

One can see this by drawing the boundingRect (outside of the transformation), in master:

master_rotated_text
Where the red dashed box is now where you have to click to select to select the text object according to the isSelected function.

By storing the transformed boundingRect, we instead get the following behaviour:
fixed_rotated_text

Which aligns with what we expect


ra3xdh:

Issue #1398, opened 2025-06-07
Segmentation fault while editing scheme


mnimonah:
qucs-s qucs s25.1.2 (637b9013)
❯ uname -a
Linux alpha 6.14.5-100.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC Fri May  2 14:22:13 UTC 2025 x86_64 GNU/Linux

Steps to reproduce:

Open BJT_Phase_Shift_Osc.sch from examplese, select all elements on canvas, Ctrl+C, create new file, Ctrl+V, try to delete any of NodeNames ("step1", "step2", "step3"...) :
systemd-coredump[3683034]: [🡕] Process 3678424 (qucs-s) of user 1000 dumped core.

ra3xdh:
I confirm this. The crash happens only after copy-paste. This may be related to #1266. @wawuwo , Please have a look.

PR #1399, opened 2025-06-08
Conductor and labels refactoring

wawuwo:
Hi, Vadim!

#1398 is once again caused by improper memory management of labels. I've refactored control of conductor labels to make it less tedious — to avoid explicit deletes, assigning nullptrs, etc. Main character here is the `Conductor` interface which is not just a plain storage for a pointer anymore, but offers a set of methods for consistent control of labels.

I've tested it a bit. It solves the problem described in #1398. I believe it should be okay overall, but can't guarantee something has not been broken. Anyway, let's give it a try, I'm here to fix in case of any problems.

Fixes: #1398

### DIFF
--- a/qucs/conductor.h
+++ b/qucs/conductor.h
@@ -1,17 +1,59 @@
 
 #ifndef CONDUCTOR_H
 #define CONDUCTOR_H
-#include "element.h"
 
-class WireLabel;
+#include "element.h"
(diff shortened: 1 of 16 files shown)


ra3xdh:
I have tested this and didn't find any issues. Merging.

PR #1403, opened 2025-06-11
Image component

andresmmera:
I've added a new component in the "Paintings" pallete to put images in the schematic.

image

It can load images from files as well as from the system clipboard. 

When the schematic is saved the image content is stored in the schematic file as base64 data.

image

**Screencast**

https://github.com/user-attachments/assets/c106e0d6-6d90-4afe-a004-a8d4aeabab88


ra3xdh:
Yes, this feature was requested by users in the past. I have planned it for the v25.2.0.

ra3xdh:
I have tested this new component and it works as expected. I have one suggestion. Apply the file extension filter when opening image files. It shows "All files". Define the allowed extensions here. 

I will merge this when the conflict will be resolved.

andresmmera:
I think that the failure in the mac-build is not related to the code:

`npm error 504 Gateway Timeout - GET https://registry.npmjs.org/@xmldom%2fxmldom`

When I try to access that URL in my web browser, I get a "504 Gateway Time-out". error.

The best solution may be to retrigger the build tomorrow and see if the issue has been resolved.

PR #1412, opened 2025-06-16
config: Add option for always prefixing datasets with the simulation label

tskaar:
This setting allows for a more consistent usage with datasets, especially in cases where one toggles on-off different simulations triggering the `a_needsPrefix` variable in NGSpice.

This setting always enforces `a_needsPrefix` giving us the dataset like `tr1.v(out)` rather than `v(vout)` in the scenario where you have one of each simulations.

There are probably other ways around fixing the "issue" at hand here, but I'd thought putting this feature behind a config flag would be the path of least resistance in terms of potentially breaking backward compatibility.

# Example

Typical use case I have seen using this:
1. Have a transient simulation
2. Have a SP-simulation
3. Make plots/diagrams from the results for both transient and SP-analysis
4. Add spectrum analysis
5. --> This triggers `a_needsPrefix = true`
6. The diagrams now don't have any valid data for the transient simulation and SP-analaysis

This config flag would then force the labeling to be `tr1.XYZ`, `sp1.XYZ` and `fft1.XYZ`, allowing the diagrams to be displayed regardless of whether one adds more analysis (triggering `a_needsPrefix`) or not

### DIFF
--- a/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/dialogs/qucssettingsdialog.cpp
@@ -147,13 +147,19 @@ QucsSettingsDialog::QucsSettingsDialog(QucsApp *parent)
     appSettingsGrid->addWidget(checkFullTraceNames, 6, 1);
     checkFullTraceNames->setChecked(QucsSettings.fullTraceName);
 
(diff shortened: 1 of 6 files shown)


ra3xdh:
Thanks for the explanations! Merged. The default setting should be `false` to ensure backward compatibility.

PR #1413, opened 2025-06-18
add CI workflow to auto-format using clang-format

dsm:
Hi @ra3xdh,

This PR adds a reusable GitHub Actions workflow that automatically formats all source files within the 'qucs' and 'qucs-*' directories using clang-format.

Key features:
- Can be manually triggered via workflow_dispatch or invoked via workflow_call from another workflow.
- Formats .cpp, .c, .h, and .hpp files.
- Creates a pull request if any changes are detected.

This helps maintain consistent code style across the codebase and reduces manual formatting overhead.

### DIFF
--- a/.github/workflows/code-format-pr.yml
+++ b/.github/workflows/code-format-pr.yml
@@ -1,11 +1,11 @@
-name: 🤖 Code Formatter & PR Bot
+name: Code Formatter
 
 on:
   workflow_dispatch:
   workflow_call:
 
 jobs:
-  format:
+  clang-format:
     runs-on: ubuntu-latest
 
     steps:


dsm:
example PR looks like this -> https://github.com/dsm/qucs_s/pull/5

ra3xdh:
> Isn't it better to run it as a check when a PR arrives?

Yes, this would be preferable, if possible.

wawuwo:
Of course this would require agreeing on style and formatting the whole codebase beforehand… But I think it's feasible and is a better approach in general.

wawuwo:
>agree on a style

clang-format is _very_ flexible. The first option is to create a bespoke style from scratch or based on one of predefined (though it probably will be very time-consuming). Another option is to simply adopt one of predefined styles — LLVM, Google, Chromium, Mozilla, WebKit, Microsoft, GNU. (WebKit is the first and GNU is the last in my personal list of favourites) Examples of these styles can be found [here](/motine/cppstylelineup). What do you guys think?

dsm:
I think we can create a custom one based off Web kit or gnu

wawuwo:
>Webkit style would be preferable for me.

Great! Then let's settle on vanilla WebKit style. I'll prepare a PR which formats all code in the repo.
In the meantime, @dsm could you please adapt this PR to introduce a check for incoming PRs as was discussed earlier?

dsm:
Hi @ra3xdh, it needs to setup a branch ruleset in repository settings to wait specific CI job success or not to block merge. I'll update to ci action that format only changed files in PR to ensure style. 

ex:
image

PR #1419, opened 2025-07-13
Fix image painting component

andresmmera:
I have found a problem with the recently added ImagePainting component: when scaling the image by dragging from the top-right corner, the image becomes significantly taller and the aspect ratio is no longer preserved.

I have also set the 'Keep Aspect Ratio' checkbox to be selected by default, as this seems to be the most convenient setting.

**Before**

https://github.com/user-attachments/assets/373c7e7d-96f9-41df-bad1-c90a37e232ab

**After**

https://github.com/user-attachments/assets/95bf5c1e-35e9-4327-b45f-88da96e1303d


ra3xdh:

PR #1426, opened 2025-07-22
Instructions on how to install using Guix

csantosb:


ra3xdh:
Thanks for pointing this, but I have to reject this PR in the current form. Here are my explanations.

I keep the installation instructions here https://github.com/ra3xdh/qucs_s/wiki/Installation . I have just added the link to this wiki page in the README. I don't see a reason to put an installation instructions for every supported distribution. I have added a common note on the wiki page in the following form:

> Qucs-S is packaged by many Linux distributions including Arch, Fedora, AltLinux, Guix, NixOS. Check this link for your repository support: https://repology.org/project/qucs-s/versions Then follow the instructions of the distribution package manager to install the binaries.

PR #1433, opened 2025-07-25
Feature: allow to move nodes

wawuwo:
Hi! I think the title says it's all, but a picture is worth a thousand words:

https://github.com/user-attachments/assets/69ab486d-691b-436e-8466-67bc46e75c17

Nodes can be moves as with mouse, and as with keyboard.

List of changes:
- refactor the way nodes store connected component and wires — use two separate containers for each specific type instead of common container for a "super" type.
- remove "lay wires anew" feature. It caused much confusion in the past and as I understand it is not very useful or loved.
- add "move nodes" feature. Right now It's on by default and cannot be turned off. Enabling "start wiring when clicking on node" overrides it completely.
- refactor here and there

I've tested it a bit on my end, seems to work fine. Anyway I'm here for a feedback and code adjustments.


ra3xdh:
Merged.

PR #1435, opened 2025-07-29
Feature: Reload netlist and log-files

tskaar:
## What

This PR adds the option for reloading already opened files, specifically netlist files and log files are being reloaded by default.

## Motiviation

When using the "Show last log" and "Show last netlist" one (usually) expects it to be the latest available netlist,
however currently on `current` if you have already opened the file, it will not refresh (getting the latest file available) and it will just show you the previously loaded netlist/log files, hence potentially giving you "stale" data. 

Typical scenario on `current` branch:
---

1. Run a simulation
2. Open up the netlist using F6
3. Leave it open
4. Change schematic
5. Run simulation
6. Open up the netlist using F6
7. --> The netlist shown is the one created during step 1.

With this PR:
---

1. Run a simulation
2. Open up the netlist using F6
3. Leave it open
4. Change schematic
5. Run simulation
6. Open up the netlist using F6
7. --> The netlist shown is now the one generated in step 5. 

## Notes

The filesystem checking etc is fairly simple in this case, since we assume that the files (netlists and log files) are purely "external", of course there are multiple ways of handling this kind of scenario.

Naturally open for suggestions/improvements if needed/wanted, thanks!

### DIFF
--- a/qucs/qucs.cpp
+++ b/qucs/qucs.cpp
@@ -1704,7 +1704,7 @@ void QucsApp::slotTextNew()
 // --------------------------------------------------------------
 // Changes to the document "Name". If already open then it goes to it
 // directly, otherwise it loads it.
-bool QucsApp::gotoPage(const QString& Name)
+bool QucsApp::gotoPage(const QString& Name, bool reloadPage)
 {
   int No = DocumentTab->currentIndex();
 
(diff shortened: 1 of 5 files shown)


ra3xdh:
Thanks! This would be a useful feature. Merged.

PR #1437, opened 2025-08-10
graphictext: Refactor code to compute boundingRect dynamically

tskaar:
What?
---
This PR refactors large parts of the code in graphictext, specifically around the handling of boundingRect.
In general this PR goes from an attribute approach to a more dynamic approach to avoid stale content.

Motivation
---
In `current`, when copying a graphictext object, one can see two effects: 

1. The boundingRect of the copy isn't necessarily rotated
2. There is an offset between where the text will be copied to and where it is previewed

Short overview
---

  1. Move transformation logic to `getTransform()`
  2. Move textBoundary logic into `getTextBounds()`
    1. This removed the discrepancy between `paint`'s and `load'`s text boundary
  3. `boundingRect` is now calculated dynamically
    1. This ensures the boundingRect is correct for both "regular" and copy-paste use
  4. `center`, `moveCenter` and `moveCenterTo` is overridden
    1. This ensures that we get the transformed center in the case where we copy-paste, removing the offset seen.
  5. Dependency/Calculations on `x2, y2, cx, cy` were removed, since we handle the boundary dynamically

Performance considerations
---
The dynamic approach creates a temporary QPixmap for text measurement when no QPainter is available. However:
- Text measurement only occurs when `boundingRect()` is called
- During painting, we reuse the existing QPainter, avoiding the temporary pixmap

I assume this performance penalty is gonna be negligible 

Other
---
This is effectively a continuation of the work started in #1397

### DIFF
--- a/qucs/paintings/graphictext.cpp
+++ b/qucs/paintings/graphictext.cpp
@@ -300,3 +300,21 @@ QRectF GraphicText::getTextBounds(QPainter* painter) const {
 
     return textBounds;
 }
+
+// Returns the center of the transformed text boundary
+QPoint GraphicText::center() const noexcept {
+    return getTransform()
+        .map(getTextBounds().center())
+        .toPoint();
+}
+
+bool GraphicText::moveCenterTo(int x, int y) noexcept {
+    QPoint currentCenter = center();
+    return moveCenter(x - currentCenter.x(), y - currentCenter.y());
+}
+
+bool GraphicText::moveCenter(int dx, int dy) noexcept {
+    x1 += dx;
+    y1 += dy;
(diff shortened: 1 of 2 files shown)


tskaar:
For reference, this is how the copy-paste issues are currently looking:

`current` - During paste operation
---
image

Showing cursor offset + no-rotation on the left graphictext

`current` - Pasted
---
image

Here showing that the boundingRect is not rotated, but the graphictext object gets rotated afterwards

With this PR - During paste operation
---
image

With this PR - Pasted
---
image

ra3xdh:
Thanks for fixing this. Merged.

Issue #1455, opened 2025-08-26
Problem with wire labels in the GUI


likecabezon:
I inserted a wire label directly on the device's pin like in the photo.

<img width="217" height="158" alt="Image" src="https://github.com/user-attachments/assets/2ca8394d-f36a-4f75-8a3c-3b8e4e92aea9" />

But when I modify the properties of the device the label moves around and connects to other device pins. The problem disappears if I connect first a wire to that terminal and attach the label to the wire.
Also I noticed that wire labels when copied they don't appear on the preview when dragging around before pasting.

ra3xdh:
I can reproduce this. This may be related to the recent redesign of schematic editor. @wawuwo , Please have a look.

PR #1457, opened 2025-08-27
Fix label changing its host after editing components properties

wawuwo:
Hi! This is fix for #1455, hope it works well.


ra3xdh:
Thanks! Merged.

PR #1459, opened 2025-08-31
Connect ComponentDialog::slotOKButton() to QDialog::accepted().

### LINKED ISSUE #1453: QT signal delivery delay when editing a component
Hi,

when I edit the properties of a component using the `Edit Properties` context-menu command like below, the changes I make are not properly saved when the dialog is closed with the keyboard.

For example, in the following dialog:

Image

if I change the capacitance value and then hit Enter (or use the shortcut Alt+O), the new value is not saved and the schematics keeps showing the old capacitance.

This never occurs when clicking on the OK button with the mouse. Besides it isn't 100% reproducible with the Enter key but it's always reproducible with the Alt+O shortcut.

I'm using the Arch Linux package (25.1.2) but it's also reproducible with the latest git head (25.2).

As I understand, this is due to a delay in the delivery of the focusOut signal in the QT library: the accept signal for QButtonBox is delivered before the focusOut signal  of the propertyTable cell.

I have to say I'm not a QT expert but it seems this is a known issue and people work around it by setting the focus on the propertyTable before saving back the changed contents.

Therefore I tried to patch the code in `qucs/components/componentdialog.cpp` as following:

```c++
void ComponentDialog::slotApplyButton()
{
...
    // ensure the table has focus to avoid data synchronization issues
    // with the focusOut signal delivery delay
    propertyTable->setFocus();
    row = 0;
    for (Property* property : component->Props)
      if ((property->simulators & QucsSettings.DefaultSimulator) != QucsSettings.DefaultSimulator) {
        continue;
      }

      // Ignore sweep parameters here.
      if (hasSweep && sweepParamWidget.contains(property->Name))
        continue;
...
```

And this solved the issue above. I'm not sure it's the right fix but it seems to work for me and solves the very annoying problem of having to change the same component twice or thrice.

funnydog:
Instead of connecting directly the QDialogButtonBox::accepted() signal to ComponentDialog::slotOKButton(), connect it to QDialog::accept() and then connect QDialog::accepted() to the slotOKButton() method.

This effectively splits the accepted signal delivery in two steps: from QDialogButtonBox to QDialog and from QDialog to slotOKButton().

On some systems, this allows the cell editors of the property table to store any pending changes before reading back the data and returning to the Qucs-S editor.

Fixes #1453.

### DIFF
--- a/qucs/components/componentdialog.cpp
+++ b/qucs/components/componentdialog.cpp
@@ -492,8 +492,9 @@ ComponentDialog::ComponentDialog(Component* schematicComponent, Schematic* schem
   QDialogButtonBox* buttonBox = new QDialogButtonBox(QDialogButtonBox::Ok | 
                                       QDialogButtonBox::Apply | QDialogButtonBox::Cancel);
 
-  connect(buttonBox, &QDialogButtonBox::accepted, this, &ComponentDialog::slotOKButton);
+  connect(buttonBox, &QDialogButtonBox::accepted, this, &QDialog::accept);
   connect(buttonBox, &QDialogButtonBox::rejected, this, &QDialog::reject);
+  connect(this, &QDialog::accepted, this, &ComponentDialog::slotOKButton);
   connect(buttonBox->button(QDialogButtonBox::Apply), &QPushButton::clicked, this, &ComponentDialog::slotApplyButton);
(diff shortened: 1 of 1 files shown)


ra3xdh:
Thanks! Merged.

PR #1486, opened 2025-10-02
Fix Monte Carlo ngspice template

ivandi69:
The template stopped working after recent changes. This PR fixes it.


ra3xdh:
Thanks! Merged.

PR #1506, opened 2025-10-29
refactor(MouseActions): extract mouse movement logic and fix tracking bug

tskaar:
This PR is mostly code cleanup and de-duplication, there shouldn't be any functional changes
other than the center boundary calculations, where component text is now disregarded to avoid (IMO) an awkward center offset.

---

What:
- Extract repeated mouse coordinate transformation and grid-snapping code into
  `updateMouseMove()` helper.
- Move element movement operations into `SchematicSelection::moveCenter()`

Why:
- Reduce code duplication.
- Make it simpler to add functionality with similar structure to mouse-movement.
- Additionally fixes a mouse-movement tracking bug where accumulation variables
  persisted across operations.

How:
- Add: `MouseActions::updateMouseMove()` that handles delta mouse movement
  calculations and tracking.
- Add: `SchematicSelection::moveCenter()` to move the logic from specific mouse
  movement functions into the struct itself.
- Add: `SchematicSelection::center()` to provide selection center coordinates.
- Fix: Check `movingState` struct rather than `movingElements`, covering a more
  generic case.
- Fix: Reset `MAx3/MAy3` in `MouseActions::MMoveMoving` so it tracks that
  particular mouse-movement operation in isolation.

### DIFF
--- a/qucs/mouseactions.cpp
+++ b/qucs/mouseactions.cpp
@@ -323,6 +323,9 @@ void MouseActions::MMoveResizePainting(Schematic *Doc, QMouseEvent *Event)
 void MouseActions::MMoveMoving(Schematic *Doc, QMouseEvent *Event)
 {
     setPainter(Doc);
+    // initialize total movement
+    MAx3 = 0;
+    MAy3 = 0;
 
     updateMouseMove(Doc, Event, /*onGrid=*/true);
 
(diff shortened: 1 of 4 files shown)


ra3xdh:
Thanks! Merged.

Issue #1510, opened 2025-11-03
Bug with microstrip filter synthesis


ra3xdh:
### Discussed in https://github.com/ra3xdh/qucs_s/discussions/1509

<div type='discussions-op-text'>

<sup>Originally posted by **psoltan** November  3, 2025</sup>
I'm trying to use the Filter Synthesis tool to create a a quarter-wave microstrip filter on FR4 dielectric (er=4.5, h=1.6mm, t=35um).  The Qucs Filter screen says it successfully calculated a filter and put it in the clipboard.  When I paste it, the microstrip dimensions are all NAN and the dielectric parameters for the substrate are all wrong.
I'm running on Linux Mint (although I had the same problem on Windows).  I've compiled Qucs-s 25.2 from source.
Is that feature broken or am I doing something wrong?

Thanks,
Phil</div>

PR #1511, opened 2025-11-03
Fix Issue #1510: Bug with microstrip filter synthesis 

andresmmera:
This PR fixes issue #1510 . The connection of the slotTakeEr() function was broken, consequently L#361:
`
    Substrate.er = ComboEr->currentText().toDouble();`

returned a wrong  _er_ and this caused the failure in the MS synthesis.

It seems that the _activated()_ signal no longer works for QComboBox.

**Before this PR**

image

**After this PR**

image


andresmmera:
Glad to help! :-)

PR #1512, opened 2025-11-05
Dev/xml devices

ThomasZecha:
-Fixed missing consideration of (rotated) symbol text bounding box in symbol layout
 for XML-based devices

### DIFF
--- a/qucs/components/xml/XmlComponent.cpp
+++ b/qucs/components/xml/XmlComponent.cpp
@@ -12,11 +12,17 @@
 #include "misc.h"
 #include "extsimkernels/spicecompat.h"
 
+#include <main.h>
+
 #include <QColor>
 #include <QPen>
 #include <QRegularExpression>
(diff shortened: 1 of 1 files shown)


ThomasZecha:
@ra3xdh: The branch is up-to-date regarding current. Any reason not to merge to current?

andresmmera:
> @andresmmera: What was your originally intentation for introducing type and angle in the XML PortSym definition, e.g. `<PortSym x="0" y="30" type="1" angle="270" />` from https://github.com/andresmmera/qucs_s/blob/bd635aae7ce9451cff0ba39457d64f3111689e67/qucs/Libraries/Sources.xml? The current C++ _Port_ definition don't have that attributes.

To be honest, I can't quite remember, but I think the _type_ field was created to number each pin. I'm not sure why I included the _angle_ field, as a pin should look the same whether it's rotated or not.

In any case, please disregard this work as an early prototype.

PR #1519, opened 2025-11-20
S-parameter Viewer: Export feature

andresmmera:
I added the Qucs-S export feature to the synthesis tool.

It converts the (internal) schematic information to a Qucs-S schematic. Some refactor was needed to make this as clean as possible.

image


ra3xdh:
Merged. I may provide feedback later if some export issues will be found.

PR #1523, opened 2025-11-24
refactor(Schematic): move node transformation logic to owner

tskaar:
## What
Remove explicit node transformation logic from the schematic layer, 
nodes are now synced through their owner elements (components/wires).

## Why
When transforming without healing, nodes ended up at incorrect coordinates because components/wires 
transformed independently of nodes. Previously, the automatic healing masked this visual error.

By moving the sync logic to owners, nodes stay consistent with their connected elements.
In the case of multi-connected nodes, they may be moved/synced multiple times, 
but due to the all node movements being based on absolute positioning, they should end up at the same coordinate. 

## How

- Components:
  - Add helper function `setNodePortCenter()` to set the node connected to that port to the same (absolute) center.
  - Call the helper function whenever we move/transform; moveCenter, rotate, mirrorX and mirrorY
- Wires:
  - Add helper function `updateP1(), updateP2(), updatePorts()` to set their respective centers to (x1, y1) and (x2, y2), updatePorts sets both.
  - Use the helper function whenever we move/transform; moveCenter, rotate, mirrorX and mirrorY
  - Use the helper function when setting a new port coordinate; i.e. in setP1 and setP2
- Schematic:
  - Remove explicit node transformations
- Schematic_selection:
  - Remove explicit node movement
  
## Demo

**Current:**

[qucs_node_transform_current.webm](https://github.com/user-attachments/assets/8a7f9d6d-dfd9-48f9-badf-c53dd1d5c4f9)

Here we apply multiple transformations in a row, so you can see that the offset keeps increasing with every transform (mirrorX & mirrorY in this case)

**With this PR:**

[qucs_node_transform_pr.webm](https://github.com/user-attachments/assets/e54d44dd-059c-4080-8699-c14be3c6b7de)

Note: The "ghost" in the middle there is a different bug, fixed in a different branch :sweat_smile:

### DIFF
--- a/qucs/components/component.cpp
+++ b/qucs/components/component.cpp
@@ -387,6 +387,18 @@ void Component::paintScheme(Schematic *p) {
         p->PostPaintEvent(_Ellipse, cx + pa->x, cy + pa->y, pa->w, pa->h);
 }
 
+bool Component::moveCenter(int dx, int dy) noexcept {
+    bool moved = Element::moveCenter(dx, dy);
+
+    if (moved) {
+        // We also need to move every port node
+        for (auto* pp : Ports) {
+            setNodePortCenter(pp);
+        }
+    }
+    return moved;
+}
+
 // -------------------------------------------------------
 // Rotates the component 90 counter-clockwise around its center
 bool Component::rotate() noexcept {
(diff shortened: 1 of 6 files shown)


ra3xdh:
Merged.

PR #1528, opened 2025-11-28
Shortcut manager

andresmmera:
I've ported the [Qucs shortcut manager](https://github.com/Qucs/qucs/pull/558), originally developed by @yodalee [here](https://github.com/yodalee/qucs/tree/feature54-custom-shortcut)

Feature requested in https://github.com/ra3xdh/qucs_s/issues/368

The dialog is accessible via File -> Application Settings

image

**Shortcut Manager dialog**
image

### DIFF
--- a/qucs/dialogs/qucsshortcutdialog.cpp
+++ b/qucs/dialogs/qucsshortcutdialog.cpp
@@ -234,73 +234,6 @@ void QucsShortcutDialog::slotChooseMenu() {
   if (actionList->rowCount() > 0) {
     actionList->setCurrentCell(0, 0);
   }
-
-  // Highlight conflicts
-  highlightConflicts();
-}
-
-// ----------------------------------------------------------------------------
-// Highlight conflicting shortcuts visually
-// The user cannot enter a duplicated shortcut, but this was introduced on an
-// early stage and it's useful to find hardcoded shortcut conflicts
-// ----------------------------------------------------------------------------
-void QucsShortcutDialog::highlightConflicts() {
-  // First, reset all backgrounds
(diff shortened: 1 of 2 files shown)


andresmmera:
So far only GND and the ports have placement actions. 

I think Qucs-S has no actions to place general components, but it good be nice to place basic components (at least resistors, caps and Ls) with a shortcut, like in LTspice.

In my opinion, this is out of the scope of this PR. It would be better to include such actions in a separate PR.

andresmmera:
Once here, I added a commit with actions for navigating the tabs:
CTRL+PageUp: Next tab
CTRL+PageDown: Previous tab
CTRL+SHIFT+PageUp: Last tab
CTRL+SHIFT+PageDown: First tab

I'm not planning to add more commit on this

yodalee:
Just want to note that, the code is written in around 2015 which is about 10 years ago. When I was still a programming newbie, I believe there will be more abstract and tidier implementation compared to mine.
Make sure you fully review the implementation before merge.

andresmmera:
I'm pushing now. I want to test the window binaries

**Main changes**

- Created a class to encapsulate the command, action ID, shortcut and default shortcut.
- Hash-based indexing. I think this is more efficient than the original implementation.
- JSON file format export. The previous implementation had a custom file format.
- Search bar.
- Keysequence is catched directly on the table. The custom LineEdit inheriting the QKeySequence is no longer needed.
- Comments in a Doxygen-like style. I think it would be a good practice to do this in general.

https://github.com/user-attachments/assets/344e619a-ebd9-421b-b761-e28b545a281c

Shahzeb-Khattak:
Looks good on my end, great work btw.

Issue #1518, opened 2025-11-19
.sym files are opened in external text editor when double-clicked in project view


jolz:
problem since version 25.2.0, version 25.1.2 worked for me.

I have en external symbol in a "*.sym"-file. on doubleclick on Projects view this is opened in an external text editor instead of open the graphical symbol. 
 
Workaround: Drag-Drop file from external file-explorer in qucs-s.

I assume, that "sym" is missing "Qucs document type list" in QucsApp::openFileFromProjectView -> https://github.com/ra3xdh/qucs_s/blob/25.2.0/qucs/qucs.cpp#L2967 so .sym is handled in fallback by mimetypes from system. (This has been added in 25.2.0, which is generally a good idea - thank you for this!).

25.1.2 had a similar condition (https://github.com/ra3xdh/qucs_s/blob/25.1.2/qucs/qucs.cpp#L2967), but i assume .sym was handled in fallback with "editfile()" https://github.com/ra3xdh/qucs_s/blob/25.1.2/qucs/qucs.cpp#L2862 an so opened internally. 

(my system/installation: Linux with xfce session, qucs-s installed via https://software.opensuse.org/package/qucs-s )

PR #1530, opened 2025-11-29
Fix issue #1518: Open sym files in Qucs-S

andresmmera:
This fixes issue #1518: if the user opens a sym file in the project editor, it's opened by the system text editor instead of being opened by the Qucs-S text editor.

The fix was already provided in the issue description.


ra3xdh:
The `*.sym` extention should be listed around line 2969 in qucs.cpp. The symbol files should be opened in schematic editor.

PR #1554, opened 2025-12-22
Add DIN A6 schematic frame

andresmmera:
I have added the DIN A6 schematic frame to the selection of available frames.

This is because the DIN A5 frame leaves too much empty space for small circuits. I often use schematics with only text or subcircuits, and I think A6 is a good size for this. A7 and smaller are very impractical.

In order to put the DIN A6 format along with the other DIN Ax sizes, I needed to add a helper function and avoid using the Combobox::currentindex directly.

An enum was added to improve code readability.

I have tested compatibility with schematics from older versions of Qucs-S and it works fine.

image


ra3xdh:
Thanks! Merged.

PR #1575, opened 2026-01-20
S-Parameter Viewer: Fixes

andresmmera:
This PR addresses the following bugs in the S-parameter Viewer:

- Markers were not being removed from the Stability and VSWR plots.
- After a dataset file was removed, the data was still present in the data map.
- After the last refactoring, the connection of the "Delete All" with the handler was broken.


ra3xdh:
Merged.

PR #1604, opened 2026-03-16
fix(AbstractSpiceKernel/getActiveLabelledNets): iteratively traverse wire nodes

### LINKED ISSUE #1603: Labelled nodes don't appear in variables list
I have found a bug related to #1594 while testing SPDT switches patch #1602. The labeled node may not appear in the output variables list. See the attahed schematic. The variables list doesn't contain the `v(in)`, but the `in` label presents on schematic. The root of this issue is hiding inside `AbstactSpiceKernel::getActiveLabelledNets` which doesn't count the `in` node as active.  I noticed this bug only for the attached schematic, but other schematics also may be affected. @tskaar , Please have a look.

[tt1.sch.gz](https://github.com/user-attachments/files/26032855/tt1.sch.gz)

Image

tskaar:
Previously only checked immediate Port1/Port2, missing wires connected in series. 
Now iteratively traverse all wires/nodes connected to components.

Example
---

image

This now properly includes `in`, which is in a (overly) complex wired setup, which wasn't the case earlier.

fixes: #1603

### DIFF
--- a/qucs/extsimkernels/abstractspicekernel.cpp
+++ b/qucs/extsimkernels/abstractspicekernel.cpp
@@ -357,19 +357,41 @@ QSet<QString> AbstractSpiceKernel::getLabelledNets(spicecompat::SpiceDialect dia
  */
 QSet<QString> AbstractSpiceKernel::getActiveLabelledNets(spicecompat::SpiceDialect dialect)
 {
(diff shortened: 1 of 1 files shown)


ra3xdh:
Thanks for providing a fix! Merged.

PR #1606, opened 2026-03-21
System Command Component

andresmmera:
This pull request (PR) introduces a new component that executes system commands once the simulation has finished. This enables the user to run post-processing in their preferred language (or any other external commands). As far as I know, Qucs-S can only run Octave scripts after the simulation has finished.

I placed the new component under the Equations tab.

image

The properties of the component are:

1. The code to execute on the shell
2. Checkbox to open a terminal emulator window
3. Checkbox to keep the emulator open

image

**IMPORTANT** I tested this on Linux (Tuxedo OS). I'll check how it works on Windows once the CI has created the corresponding binaries.

**Example**

In the example attached, a LPF with finite Q components (qucsator) is simulated and a system-command block is added to compare the results with a simulation in scitkit-rf (ideal components)

[SystemCommandFeatureTest_prj.zip](https://github.com/user-attachments/files/26156888/SystemCommandFeatureTest_prj.zip)

image

Python output
image

### DIFF
--- a/examples/Postprocessing/ExternalPostprocessing_prj/LPF.sch
+++ b/examples/Postprocessing/ExternalPostprocessing_prj/LPF.sch
@@ -1,0 +1,49 @@
+<Qucs Schematic 25.2.0>
+<Properties>
+  <View=-142,-165,1644,829,0.880639,0,0>
+  <Grid=10,10,1>
+  <DataSet=LPF.dat>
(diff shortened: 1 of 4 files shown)


andresmmera:
Thanks for your review.
I'm currently working on the highlight feature. I have resolved the issue with the selection box.
I'll push the fix soon.

PR #1628, opened 2026-04-13
fix(port): handle empty case (terminated port)

tskaar:
## What

This allows to netlist ports when power is unset.
The assumption is that the user then wants a terminated (passive) load, i.e. we set the AC (and transient) voltage to 0 V.

## Why

Since the changes #1617, if we have an unset power parameter, we hit the "symbolic" branch of the power calculation, giving us an expression where we are dependent on the power value/parameter.

However, that is unset in this particular case, causing a netlisting error.

## Example

Take this simple testbench:

image

Where P1 has a set power, and P2 has an unset power.

### Current

If I try to netlist this using NGSpice I end up with:

```
VP1 _net0 0 dc 0 ac 2 portnum 1 z0 50
VP2 _net1 0 dc 0 ac '2*sqrt(2)*sqrt(50/1000)*pow(10,()/20)' portnum 2 z0 50
                                                    ^^^
                                           Empty pow parameter
```

### This PR

```
VP1 _net0 0 dc 0 ac 2 portnum 1 z0 50
VP2 _net1 0 dc 0 ac 0 portnum 2 z0 50
```

Here I am explicitly setting AC voltage to 0 V (and transient voltage if that was part of the schematic), and NGSpice can run on this netlist.

Note: Also verified the behavior on Xyce

### DIFF
--- a/qucs/components/source_ac.cpp
+++ b/qucs/components/source_ac.cpp
@@ -71,6 +71,8 @@ Source_ac::Source_ac()
         QObject::tr("simulation temperature in degree Celsius")));
   Props.append(new Property("EnableTran", "true", false,
     QObject::tr("enable transient model as sine source [true,false]")));
+  Props.append(new Property("LoadOnly", "false", false,
+    QObject::tr("disable as a source (AC and transient), passive termination only [true,false]")));
 
   rotate();  // fix historical flaw
 }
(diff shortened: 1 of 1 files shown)


andresmmera:
Oh, I missed that case! 
I’ve confirmed the bug, and the fix works as expected.

ra3xdh:
I didn't know about this implicit feature. Looks like undefined behavior. I suggest to introduce a more straightforward way for setting source magnitude to zero. I would propose to introduce an explicit `LoadOnly=yes/no` property (or choose another name). The default value is `no`. If it is set to `yes`, the power value defined by `P` is ignored and the netlist entry will have `AC 0`. New properties must be added to the end of the properties list. Otherwise the load of schematics from previous version will be broken.

tskaar:
Sure, a more explicit flag wouldn't hurt. I can make the changes there.

Any preference for handling the empty case in general, since it currently triggers a netlisting error?

ra3xdh:
Empty string could be handled as implemented by this commit resulting as `AC 0`

tskaar:
Pushed the changes now, where I now account for `LoadOnly`:

image

ra3xdh:
Merged.

PR #1634, opened 2026-04-15
fix(SpiceLibComp): pin parsing and dialog input validation

### LINKED ISSUE #1632: SpiceLib: Continuation line (`+`) in subckt definition is not taken into account
Hi,

When trying to import a netlist that has ports over multiple lines, i.e. it uses continuation lines (`+`), these are ports are not taken into account.
From what I can tell, it only parses the initial SUBCKT line and then skips the rest.

## Reproducer

As a simple example I made the following subckt definition:

```
.SUBCKT RLadder VDD
+ TAP1
+ TAP2 TAP3
+ GND
	R1 VDD  TAP1 10k
	R2 TAP1 TAP2 10k
	R3 TAP2 TAP3 10k
	R4 TAP3 GND  20k
.ENDS
```

This subckt is intentionally split over multiple lines to check that it parses it as expected.

And when instantiating this using the `SpiceLib` component I get the following:

Image

Where only VDD is populated, which is the only port/signal that is part of the initial subckt definition line

## TODO

I'll submit a fix for this

tskaar:
## What

- `spicecompat::getPins`: Fixes incorrect pin counting when subcircuits span multiple lines.
- `SpiceLibComp::spice_netlist`: input sanitation and validation, to handle malformed dialog input. 

## Why

- `getPins` previously only updated the pin count based on the initial definition line, even though all pins were correctly added to the list. This caused a mismatch between the reported pin count and the actual list of pins. 
- `spice_netlist` triggered a crash (SIGABRT) when encountering empty or invalid pin inputs from the dialog. Duplicate separators caused the split logic to produce empty strings, leading to an out-of-bounds access (`Ports.at(-1)`)

## How

- `getPins`: Switched to incrementing the pin counter on every append.

- `SpiceLibComp::spice_netlist`:
  - Split input using `Qt::SkipEmptyParts` to avoid duplicated separators
  - Added bounds checking, and type validation before trying to access `Ports`.
  - Added logging through `qWarning` in case we see any "malformed" inputs.
  
## Example

Using the same setup as in #1632, it now parses the ports correctly and we now get:

image

  
  
fixes: #1632

### DIFF
--- a/qucs/extsimkernels/spicecompat.cpp
+++ b/qucs/extsimkernels/spicecompat.cpp
@@ -302,6 +302,7 @@ int spicecompat::getPins(const QString &file, const QString &compname, QStringLi
                 break;
               } else {
                 pin_names.append(pin);
+                r++;
               }
             }
           } else {
@@ -325,10 +326,9 @@ int spicecompat::getPins(const QString &file, const QString &compname, QStringLi
                 if (!s1.contains('=') &&
                    (pp.toLower() != "params:")) {
                     pin_names.append(s1);
+                    r++;
                 }
             }
-            r = pin_names.count();
-            //break;
         }
     }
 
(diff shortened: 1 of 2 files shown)


ra3xdh:
Thanks! Merged.

PR #1644, opened 2026-04-26
feat(frames): Add large frames; A2, A1 and A0

tskaar:
## What

This adds DIN A{2,1,0} frames

## Why

I have some massive subcircuits, where A3 was too small...

## How

The dimensions were derived by multiplying the previous format's dimensions by sqrt(2) (approx. 1.5) and rounding up to the nearest column/row step size (to avoid partial columns/rows at the frame edge), giving: 

- (already implemented) `A3_landscape`: 9 columns
- `A2_landscape`:  13 columns
- `A1_landscape`:  20 columns
- `A0_landscape`: 30 columns

**Refactor**

Refactored `addFrameItem` helper function such that we pass `FrameSize` enum values directly instead of manually specifying raw integer values.


ra3xdh:
Thanks! Merged.

PR #1646, opened 2026-04-27
System Command component: Add terminal emulator field

andresmmera:
On Linux, when a system command component is used, Qucs-S searches a hardcoded list of terminal emulators and uses the first one available [1]. In Lubuntu and Fedora, this fails because the terminal emulator shipped with those distros are not in the list:

image

This PR adds a new property field to specify which terminal emulator to use. If left blank, Qucs-S falls back to trying the hardcoded list, as before.

image

Two terminal emulators were added to the default hardcoded search list:

- Qterminal, shipped with Lubuntu
- Ptyxis, shipped with Fedora

[1] https://github.com/ra3xdh/qucs_s/blob/current/qucs/qucs.cpp#L4048


ra3xdh:

PR #1652, opened 2026-05-09
Fix Shortcut Manager persistance

andresmmera:
As reported here https://github.com/ra3xdh/qucs_s/discussions/1651, the shortcuts settings do not hold for the next Qucs-S session.

After inspecting the code, Qucs-S is not loading the user shortcuts from qucs_s.conf. Once this was fixed, it was observed that it didn't save the shortcuts when they're reseted to default.

**Fixes**

- Call loadFromSettings() at the end of setDefaultShortcut(), after all commands have been registered, so saved overrides are applied correctly on startup

- Remove QSettings keys for commands reset to default, preventing stale overrides from being reapplied on next launch


ra3xdh:
Thanks for providing fix! Merged.

andresmmera:
Thanks!

PR #1662, opened 2026-05-19
Fix Xyce HB for multi-tone simulations

andresmmera:
A colleague reported to me that multitone HB simulations do not work with the Xyce backend. This happens because *numfreqs* requires the same number of arguments as the HB frequencies. (See [Xyce Reference Guide,](https://xyce.sandia.gov/documentation-tutorials/) section 2.1.25.17)

image

**Example**

image

image

This is how the Xyce netlist looks like now:
image

After this PR

image

Now it simulates without errors:

image

It works with N carriers
image

Here it is the workspace I used for testing this: [Xyce_Multitone_HB_prj.zip](https://github.com/user-attachments/files/28024844/Xyce_Multitone_HB_prj.zip)
I need to polish somethings on this project. Then, I think I would be worthy of adding it to the examples.


ra3xdh:
Thanks! Merged.

PR #1664, opened 2026-05-25
Fix Xyce parametric transient sweep misalignment in parseSTEPOutput()

### LINKED ISSUE #1663: Data misalignment in parametric transient simulations with Xyce
I’ve noticed that **_parametric_** transient simulations in Xyce produce misaligned data depending on the number of points. If I perform the simulations separately (i.e. without the sweep block), they are ok.

I think the problem lies in how Qucs-S parses the Xyce dataset. When  `~/.cache/qucs-s/spice4qucs.tran._swp-plot` is opened with the ngspice viewer, it shows ok, but the same data on Qucs-S is misaligned.

I suspect the problem lies in this function https://github.com/ra3xdh/qucs_s/blob/current/qucs/extsimkernels/abstractspicekernel.cpp#L872. I also suspect that the time vector in Xyce is not identical in each iteration. 

I’m trying to fix this, but I haven’t found a solution yet. Any insight is appreciated.

Here’s a schematic where the parametric transient simulation in Xyce fails. [PowerSweep_xyce.sch.zip](https://github.com/user-attachments/files/28113200/PowerSweep_xyce.sch.zip)

Image

**Xyce raw data**

Image

**Xyce data plotted with the ngspice viewer**
Image

andresmmera:
This PR fixes #1663 

**Problem**

Xyce transient simulator has a adaptive timestep that produces different time vectors in each sweep step. The function convertToQucsData() assumes a single time vector for all the time series and, consequently, transient parametric simulations in Xyce show misaligned waveforms when displaying the results in Qucs-S.

As discussed in #1663, this is a limitation of the Qucs dataset format, but interpolating the time series with respect to the same time vector would solve the problem.

**Fix details**

**1. Collect sweep steps separately during parsing**
Before this PR, *No. Points* was only read from the first block header (guarded by header_parsed). Subsequent block headers were ignored, so there was no way to detect where one sweep step ended and the next began. 
The fix moves *No. Points* outside the header_parsed guard so it is re-read for every block, and uses it as the trigger to save the completed step and start a new one. 
This way each sweep step is accumulated separately before being passed to resampleToCommonGrid().

**2. Resample onto a common uniform time grid**
A new function *resampleToCommonGrid()* was added to interpolate all steps onto a common uniform time grid before merging into sim_points. Resampling is only applied when the independent variable is "time". 

**Comments**

AC, DC and other sweep types are unaffected and follow the original code path.

I've used the following project to ensure this PR fixes the problem and doesn't cause trouble to other simulations. Specifically, I tested DC, AC, SP and TR simulations with and without parametric sweeps (both ngspice and xyce).

[XycePowerSweepBug_prj.zip](https://github.com/user-attachments/files/28204544/XycePowerSweepBug_prj.zip)

image

### DIFF
--- a/qucs/extsimkernels/abstractspicekernel.cpp
+++ b/qucs/extsimkernels/abstractspicekernel.cpp
@@ -891,6 +891,12 @@ void AbstractSpiceKernel::parseSTEPOutput(QString ngspice_file,
     bool header_parsed = false;
     int NumVars=0; // Number of dep. and indep.variables
     int NumPoints=0; // Number of simulation points
+
+    // Per-step storage for resampling (Xyce adaptive timestepper produces
+    // variable-length steps; we interpolate onto a common grid before merging)
+    QList< QList< QList<double> > > all_steps;
+    QList< QList<double> > current_step;
+
     while (!ngsp_data.atEnd()) {
         QRegularExpression sep("[ \t,]");
         QString lin = ngsp_data.readLine();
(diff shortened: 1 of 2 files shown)


ra3xdh:
Thanks for proving a fix. Merged.

PR #1666, opened 2026-05-25
feat(SpiceOptions): add Xyce option package dropdown

tskaar:
##  What

Adds an editable combobox to the SpiceOptions dialog for Xyce option packages, pre-filled with all packages from Xyce v7.10 reference guide, with descriptions. Valid selections show a green checkmark/border; invalid ones show a yellow warning (Xyce silently skips invalid packages). The control is hidden/disabled when using non-Xyce simulators.

Also fixes `hasDarkTheme` cache not updating on style change in QucsSettingsDialog.

## Why

Currently, it feels a bit weird having to have the XycePkgOption enforced as the first line, even when using NGSpice, hence I felt this would be a more consistent experience (in the case when using Xyce), for NGSpice/Non-Xyce users, the change should be invisible. 

NOTE: This should not break any pre-existing options, since one is using the same parameters as previously used.

### Schematic Entry

qucs_options_schematic

Where one can see that the `XyceOptionPackage` is now hidden for NGSpice.

### Dialog

| NGSpice | Xyce |
|-----------|-------|
| image |  image |

Showing that NGSpice disables the combo-box, while Xyce is currently showing an invalid option

### Dropdown options

image

### Input validation

| Valid | Invalid/Unknown |
|-------|--------------------|
| image |  image | 

### Light vs Dark Theme

|   Input  | Adwaita (light) | Adwaita-dark |
|---------|------------------|----------------|
| Invalid | image| image |
| Valid    |image | image |

Thanks!

### DIFF
--- a/qucs/components/componentdialog.cpp
+++ b/qucs/components/componentdialog.cpp
@@ -35,6 +35,8 @@
 #include "misc.h"
 #include "fillfromspicedialog.h"
 #include "selfromlibdialog.h"
+// For Xyce Options
+#include "spicecomponents/sp_options.h"
 
 #include <cmath>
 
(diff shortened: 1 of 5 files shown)


ra3xdh:

PR #1669, opened 2026-05-29
Multitone AC power sources

andresmmera:
This PR extends the AC power source component to handle multiple frequencies and phases. The goal is to simplify circuit setup for multitone simulations. The AC power source component now accept a list of frequencies and phases and construct the netlist accordingly.

**Two-tone test before this PR**

image

**Two-tone test with this PR**

image

Once working on this, I’ve found that the qucsator-rf power source give wrong results in HB and transient simulations. In order to fix this I replaced the qucsator-rf power source instantiation by an AC voltage source + a Z0 resistor in the netlist. Now the qucsator-rf results are consistent with Xyce (HB and tran) and ngspice (tran)

This is the workspace I've used for testing this new features. [Multitone_AC_Power_prj.zip](https://github.com/user-attachments/files/28377436/Multitone_AC_Power_prj.zip)

Tests:
    • Single-tone and Multitone HB: Verified on both qucsator-rf and Xyce backends. Amplitude and phase match expected results. 
    • Single-tone transient: Results are consistent across all three backends (qucsator-rf, Xyce, ngspice). The phase parameter is ok.
    • Multitone Transient Simulations: Waveforms are consistent across all simulation backends. 

image

image

### DIFF
--- a/qucs/components/hb_sim.cpp
+++ b/qucs/components/hb_sim.cpp
@@ -90,3 +90,22 @@ QString HB_Sim::spice_netlist(spicecompat::SpiceDialect dialect /* = spicecompat
     }
     return s;
 }
+
+QString HB_Sim::netlist()
+{
+  QString fVal = Props.at(0)->Value.trimmed();
+
+  // qucsator reads tone frequencies directly from PAC excitation components.
+  // Its f property is PROP_REAL and cannot parse a comma-separated list —
+  // omit it when multiple tones are specified; the PAC chain handles it.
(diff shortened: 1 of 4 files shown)


ra3xdh:
Thanks for adding new feature! Merged.

andresmmera:
Glad to help. Thanks!

andresmmera:
I suspect the S-parameter simulation is broken (because of this).
I'll take a look at it this afternoon.

PR #1671, opened 2026-06-02
fix(spiceLibComp): handle non-linear port indices

### LINKED ISSUE #1668: SpiceLib(custom symbol): assumes a linear port range, causing missing/incorrect ports
Hi,

I noticed that when instantiating a `spiceLibComp` with a custom symbol, the dialog assumes that ports are defined in a linear range starting from index 1 and going up to `numPorts`.
If the symbol has non-linear or sparse port numbers, the dialog skips over missing ports and incorrectly lists "empty" or out-of-bounds ports

E.g. if a symbol has the following ports:

| PortNum | PortName |
|:----------:|:------------|
| 1             | P1              |  
| 3             | P3              |
| 4             | P4              |
| 5             | P5              |

The dialog would then populate the following:

Image

The dialog incorrectly populates the port list, showing port 2 (which doesn't exist) and not getting to P5 (out of bounds relative to `numPorts`)

## Re-producer

1. Use a netlist (I used the same as in #1632):

```
.SUBCKT RLadder VDD
+ TAP1
+ TAP2 TAP3
+ GND
	R1 VDD  TAP1 10k
	R2 TAP1 TAP2 10k
	R3 TAP2 TAP3 10k
	R4 TAP3 GND  20k
.ENDS
```

2. Create a schematic with unordered/non-linear ports, e.g.:

Image

3. And then save that schematic as a symbol only, in my case:

```
<Qucs Schematic 26.1.1>
<Symbol>
  <.ID -20 44 SUB>
  <.PortSym -30 -30 1 0 P1>
  <.PortSym 30 -30 3 180 P3>
  <.PortSym -30 30 4 0 P4>
  <.PortSym 30 30 5 180 P5>
  <Line -30 -30 10 0 #000080 2 1>
  <Line 20 -30 10 0 #000080 2 1>
  <Line -30 30 10 0 #000080 2 1>
  <Line 20 30 10 0 #000080 2 1>
  <Line -20 -40 40 0 #000080 2 1>
  <Line 20 -40 0 80 #000080 2 1>
  <Line -20 40 40 0 #000080 2 1>
  <Line -20 -40 0 80 #000080 2 1>
</Symbol>
```

4. Finally, instantiate a `spiceLibComp` block, and the populate your netlist and custom symbol
(see first image in the post), and now check that all your ports are listed as expected.

**Expected behaviour**

All ports defined in the symbol should appear in the dialog, regardless of their indices.

## TODO

I'll post a fix for this

tskaar:
### What

Use actual port indices from `PortNames`  instead of assuming a linear range. 
Additionally, a `qWarning` is added when overriding a port (same index/number). 
For netlisting, pin numbers are normalized to ranks to handle non-linear or sparse port indices.

### Example:

Using an extended example from #1668, where ports are defined as (index, name): (1, P1), (3, P3), (4, P4), (5, P5), (6, P6),
the dialog now currently displays:

image

Where we now can see that the second port is now P3 and not an empty field.

**Netlisting Example**

With the following map:

image

I get the netlist:

```
* Qucs 26.1.1  /home/torleif/qucs_test.sch
.INCLUDE "/home/torleif/qucs_bug/spicelib_iterator/subckt.cir"
XX1  SYM4 SYM3 SYM1 SYM5 SYM6 RLADDER 
.END
```

Which I would consider _correct_ in this case

**NOTE:**

Since every port must be mapped, ranked indexing works, but it would break if ports could be arbitrarily skipped (would require netlisting to a floating net).

Fixes: #1668

### DIFF
--- a/qucs/extsimkernels/spicelibcompdialog.cpp
+++ b/qucs/extsimkernels/spicelibcompdialog.cpp
@@ -444,7 +444,10 @@ void SpiceLibCompDialog::slotTableCellDoubleClick()
 
   QComboBox *cbxSelectPin = new QComboBox;
   cbxSelectPin->addItem("NC");
-  for (int i = 1; i <= a_symbolPinsCount; i++) {
+  // Ports might not be in a linear range; use the actual indices directly
+  // to ensure correct lookup
+  QList<int> portIndices = a_symbol->getPortIndices();
+  for (int i : portIndices) {
     bool pinAssigned = false;
     for(int j = 0; j < a_tbwPinsTable->rowCount(); j++) {
       if (j == r) continue;
--- a/qucs/spicecomponents/spicelibcomp.cpp
+++ b/qucs/spicecomponents/spicelibcomp.cpp
@@ -232,25 +232,52 @@ QString SpiceLibComp::spice_netlist(spicecompat::SpiceDialect dialect /* = spice
   } else {
     // Sanitize input from dialog
     QStringList pin_nums = pins.split(";", Qt::SkipEmptyParts);
-    int pinCount = pin_nums.count();
     int portCount = Ports.count();
 
-    if (pinCount != portCount) {
-      qWarning() << "pinCount=" << pinCount << " != portCount=" << portCount << ", malformed input?";
+    // Extract port numbers and check for duplicates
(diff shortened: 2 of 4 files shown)


ra3xdh:
Merged.

PR #1672, opened 2026-06-08
Fix Smith Chart: Single point data representation fails with "solid line" linestyle

andresmmera:
**Problem**

The Smith Chart diagram is empty when the S11 is constant and the "solid line" linestyle is selected.

Lines are drawn  in Graph::drawLines(QPainter* painter). This condition https://github.com/ra3xdh/qucs_s/blob/current/qucs/diagrams/graph.cpp#L559 is used to decide whether paint a point or not.  When the data is constant, all points map to the same coordinate, so no segment is ever drawn.

image

**Fix**
Once the loop is finished, check if "lines" is empty and if so, fall back to a star plot.

image

**Additional Changes**

While working on this, some warnings were spotted and fixed.

### DIFF
--- a/qucs/diagrams/graph.cpp
+++ b/qucs/diagrams/graph.cpp
@@ -111,7 +111,7 @@ QString Graph::save()
 	      " "+QString::number(numMode)+" "+QString::number(Style)+
 	      " "+QString::number(yAxisNo)+">";
 
-  for (Marker *pm : Markers)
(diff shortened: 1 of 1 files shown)


ra3xdh:
Yes, it is known problem. The Par.Sweep has similar problem if the graph consists of 1x1 point plots on every step.  Thanks for providing a fix.

PR #1674, opened 2026-06-11
Fix QucsatorRF simulation regressions after #1669 (SP, power sweep, backward compatibility)

andresmmera:
**Problem**

After merging PR #1669, the following issues were identified:

**1) Qucsator-RF SP simulation is broken**
PR #1669 modified the qucsator's netlist generation for AC Power sources to use a constant AC Voltage source + Z0 series resistor. This was intended to be done in HB and TRAN simulations only, but it was accidentally applied to SP simulations as well.

image

**2) Backward compatibility broken for old schematics**
PR #1669 added the "phase" property to the AC power source component. Since Qucs-S loads component properties positionally from the schematic file, any schematic created before this change loads with shifted property values: "Phase" receives the "temperature" field. 
Positional loading seems to be a bad practice, hopefully it'll be fixed by the IHP team with the new XML component schema, #1354 

**3) Broken HB/TRAN power sweep in qucsator-RF**
3.1) After #1669, the peak voltage amplitude is computed from the available power at netlist generation time. When the power property contains a sweep parameter (e.g. Pin) rather than a numeric value, the conversion fails silently and the amplitude is fixed at zero for all sweep steps.

image

3.2) For linear dBm sweeps, QucsatorRF converts the start/stop endpoints from dBm to watts and interpolates linearly in watts, producing non-uniform dBm spacing instead of the intended uniform dBm steps.

---
**Fixes**

**1)** The Pac component need to know which type of simulation the user wants to run. For doing this, a new field was introduced in Schematic class to identify the simulation before each component generates its portion of netlist. 

Commits:
https://github.com/ra3xdh/qucs_s/commit/1cd79b61fb66786eb8e1676347a6c2df53b0a846
https://github.com/ra3xdh/qucs_s/commit/f10ce7c7db70673690e076b20e3512daf16155bd

**2)** Automatically add the phase property to Pac and Vac sources if it's not present at loading. If they don't contain enough property fields, it means that the schematic don't include the phase property.

Commit:
https://github.com/ra3xdh/qucs_s/commit/22247052a64ac147f2d8b4f3ca015fc2c64891b5

**3.1)** When the power value is non-numeric, a companion Eqn: block is emitted alongside the Vac: source to compute the amplitude symbolically at each sweep iteration.
Commit:
https://github.com/ra3xdh/qucs_s/commit/c368f11cfa4a5cde367c4099bd33194b0bef0f21

**3.2)**  Overload Param_Sweep::netlist() by mapping linear dBm sweep into an explicit list of watts values, preserving the correct dBm spacing at each step.
https://github.com/ra3xdh/qucs_s/commit/5420809befc7e03a69d56a57423381f039de5e94

*Note: Fixed some QString multiarg warnings once there.*

I attach the schematics I used for testing all this: [Multitone_AC_Power_prj.zip](https://github.com/user-attachments/files/28821597/Multitone_AC_Power_prj.zip)

### DIFF
--- a/qucs/components/component.cpp
+++ b/qucs/components/component.cpp
@@ -1818,6 +1818,31 @@ Component *getComponentFromName(QString &Line, Schematic *p) {
         return 0;
     }
 
+    // Migrate old Pac/Vac schematics saved before Phase property was inserted at index 4 (PR #1669).
+    // Old order (7 props): Num, Z, P, f, Temp, EnableTran, LoadOnly
+    // New order (8 props): Num, Z, P, f, Phase, Temp, EnableTran, LoadOnly
+    // In old versions, Temp contains a boolean string because old positional values shifted, so the
+    // idea is to detect that and fix it here.
+    if (c->Model == "Pac" || c->Model == "Vac") {
+      Property *pTemp = c->getProperty("Temp");
+      if (pTemp && (pTemp->Value == "true" || pTemp->Value == "false")) {
+        // Values are shifted: index 4 (Phase) has Temp value, index 5 (Temp) has EnableTran value
+        Property *pPhase = c->getProperty("Phase");
+        Property *pEnableTran = c->getProperty("EnableTran");
+        Property *pLoadOnly = c->getProperty("LoadOnly");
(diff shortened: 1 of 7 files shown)


andresmmera:
Thanks

Issue #1692, opened 2026-07-23
Automatically place DC bias labels to avoid schematic overlaps


viticleri:
## Problem

When DC bias points are displayed, voltage and current labels can overlap
schematic elements such as components, probe symbols, node markers, wires,
component text, or other bias labels. This makes dense schematics harder to
read even when the simulation results are correct.

Internally, both voltage and current bias labels are positioned from node
coordinates, using fixed offsets. A fixed offset cannot adapt reliably to
different component orientations or local schematic density.

## Suggested improvement

Add automatic placement for DC bias labels.

For each label, Qucs-S could:

1. Generate several candidate positions around its anchor point, for example
   upper-left, upper-right, lower-left, and lower-right.
2. Calculate the label rectangle for each candidate.
3. Score or reject candidates that intersect:
   - component bounding rectangles;
   - node markers;
   - wire segments;
   - component labels or property text;
   - DC bias labels already placed.
4. Choose the candidate with the fewest conflicts and a reasonable distance
   from the anchor.
5. Use a defined fallback when no collision-free position exists, such as the
   least-overlapping candidate or a small leader line.

The label should remain visually associated with the correct node, current
probe, or source.

## Expected result

- Voltage and current bias labels are placed beside their anchor instead of
  directly over schematic objects whenever space is available.
- The behavior works consistently for rotated and mirrored components.
- Existing DC values, units, colors, and simulation behavior remain unchanged.
- Dense schematics degrade gracefully when a completely collision-free
  position is impossible.

## Scope note

This is intentionally broader than issue #1316. That issue addresses one
specific overlap involving voltage-probe terminals. This request proposes a
general label-placement strategy and should be implemented and reviewed
separately.

## Possible acceptance criteria

- A placement helper can be tested independently with synthetic rectangles and
  wire segments.
- Tests cover free placement, component overlap, wire overlap, label-to-label
  overlap, and the no-free-position fallback.
- Manual verification includes a dense schematic with voltage and current bias
  labels.

PR #1693, opened 2026-07-23
Avoid DC bias labels over voltage probes

### LINKED ISSUE #1316: Wrong dc bias value calculated in Votalge probe.
Hello, good afternoon.

Im seeing wrong dc bias calculation in voltage probe while measure voltage drop in a 1n4148 diode.
Value drop is correct in table.

Image

Wrong value in voltage probe bias calculation.

Image

viticleri:
## Summary

- suppress a DC bias label when its rectangle overlaps a directly connected
  voltage probe;
- keep DC bias labels unchanged for other component types and for voltage
  probes that do not overlap the label;
- add focused tests for the probe-model and rectangle-intersection decision.

## Problem

DC bias labels are positioned relative to node coordinates. At voltage-probe
terminals, the voltage labels can overlap each other and cover the probe
symbol, making values such as `5V` and `0V` appear visually combined.

The calculated DC values are correct; this change only addresses their
rendering.

## Implementation

The renderer calculates the label rectangle before drawing it and checks the
components directly connected to the node.

The label is skipped only when:

- the connected component model is exactly `VProbe`; and
- the label rectangle intersects the component bounding rectangle.

This avoids suppressing labels merely because a voltage probe is connected to
the node.

## Testing

  - `DcBiasLabelTest`: passed
  - `GeometryTest`: passed
- manual verification with the reproduction schematic:
  - the labels no longer cover the `VProbe` terminals or symbol;
  - voltage and current labels elsewhere remain visible.

Fixes #1316

A broader automatic placement strategy for DC bias labels is tracked
separately in #1692.

### DIFF
--- a/qucs/CMakeLists.txt
+++ b/qucs/CMakeLists.txt
@@ -149,11 +149,13 @@ SET(QUCS_SRCS
   imagewriter.cpp printerwriter.cpp projectView.cpp
   symbolwidget.cpp wire_planner.cpp
   qucsshortcutmanager.cpp
+  dcbiaslabel.cpp
 )
 
 SET(QUCS_HDRS
 element.h
 conductor.h
+dcbiaslabel.h
 healer.h
 main.h
 messagedock.h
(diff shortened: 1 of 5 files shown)


ra3xdh:
The solution provided by this patch is overcomplicated. I have prepared a simpler patch based on the proposed idea. See #1694 Thanks for contribution and closing this.

viticleri:
Thank you for reviewing the proposal and for preparing and merging the simpler fix in #1694. I’m glad the contribution helped resolve #1316.

PR #1698, opened 2026-07-28
Inline document rename

andresmmera:
This PR adds the ability to rename open documents, either by double-clicking the tab name or via the "Rename" entry in the tab's context menu. Previously, Qucs-S couldn't rename open files (even though this was technically possible). This PR fixes that.

When the document is open, its name is edited directly on the tab: a QLineEdit is temporarily created and placed there. If the file to rename isn't open, it can still be renamed from the project panel, as before, preserving the existing functionality.

The new renameDocumentTab() function shares some logic with the existing slotCMenuRename(). This shared logic has been moved into QucsApp::renameFileOnDisk(), which both functions now use, avoiding duplication.

**Rename action added to tab context menu**
image

**Screencast about the inline document rename feature**
[InlineRename.webm](https://github.com/user-attachments/assets/f4de82bc-eae9-4132-86e6-1ec946ebdecd)

### DIFF
--- a/qucs/qucs.cpp
+++ b/qucs/qucs.cpp
@@ -1326,6 +1326,17 @@ void QucsApp::slotCMenuRename()
   QString file(QucsSettings.QucsWorkDir.filePath(filename));
   QFileInfo fileinfo(file);
 
+  int index=0;
+  QucsDoc* Doc = findDoc(file, &index);
+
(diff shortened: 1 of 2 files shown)


ra3xdh:
Thanks! Merged.

PR #1700, opened 2026-08-02
Add pre-simulation schematic checks

### LINKED ISSUE #1697: [Subckts] Display a warning in the case of non-connected ports and dangling nets
When using subcircuits, often times I get silent errors in the case where I have non-connected pins, e.g. I have a typo between my labelled net and my port-net. 

Displaying a warning (or some other visual effect) stating that there are non-connected ports would significantly help with getting rid of these issues early on.

## Example

if I have the subcircuit:

Image

In this case Port `P1` is not attached to anything, additionally `P1_net`, `P1` and `P2` are all dangling, so it might be worth also to warn the user that these dangling nets might have unforeseen effects when doing the simulation.

## TODOs

- The user might not want the additional "noise" here, so it might be worth hiding this option behind a configuration flag.

## Contribution

I'll work on a PR for this

Related: #1696, this is essentially "part II" of this issue

andresmmera:
**Motivation**
Sometimes simulations fail because small mistakes that may be hard to read in the simulator error log.
 
This PR introduces a pre-simulation schematic check for common issues. The goal is to catch common configuration issues early, before the simulator crashes.

Features:

- Checks are propagated to the subcircuits.
- It runs only in the single-run simulation mode. It doesn't make sense to run the checks in tuning mode and, it would also be a mess.
- The user can still run the simulation even if the checks found errors.
- Three categories: "Critical" (the simulation can't run, major errors), "Warning" (something is odd, but it may run), "Minor" (dangling wires, etc.)
- This feature can be disabled in the Qucs-S Settings panel.

image

The errors/warnings (if any) are listed in a pop-up window.

image

**Checks**

1. List-based frequency sweep
Validates that the SP/AC frequency sweep is not defined as a list when the simulation backend is ngspice or xyce. 
qucsator supports list-based frequency sweeps, but ngspice and xyce do not. Previously, switching backends from qucsator to ngspice/xyce on a schematic with a list-based sweep would cause the simulation to fail with an unhelpful error.

2. Minimum port count for ngspice SP simulation
ngspice requires at least two ports to run an SP simulation. Added a check to catch schematics with fewer than two ports before simulation is attempted.

3. Missing simulation block
All backends require at least one simulation block to be present. xyce in particular fails silently when none is defined.

4. Propagate checks to subcircuits. This is useful to fix this: https://github.com/ra3xdh/qucs_s/issues/1697

6. Dangling wires and nets connected to nothing
This is marked as minor, as it shouldn't break the simulation

8. Check for unconnected ports

This is the workspace I used for checking this feature
[SchematicValidation_prj.tar.gz](https://github.com/user-attachments/files/30575032/SchematicValidation_prj.tar.gz)

@tskaar I added a test to catch dangling wires and unconnected ports. Please check if this fixes https://github.com/ra3xdh/qucs_s/issues/1697

### DIFF
--- a/qucs/CMakeLists.txt
+++ b/qucs/CMakeLists.txt
@@ -149,6 +149,7 @@ SET(QUCS_SRCS
   imagewriter.cpp printerwriter.cpp projectView.cpp
   symbolwidget.cpp wire_planner.cpp
   qucsshortcutmanager.cpp
+  schematicvalidator.cpp
 )
 
 SET(QUCS_HDRS
@@ -166,6 +167,7 @@ octave_window.h
 qucs.h
 qucsdoc.h
 qucsshortcutmanager.h
+schematicvalidator.h
 schematic.h
 settings.h
 syntax.h
--- a/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/dialogs/qucssettingsdialog.cpp
@@ -164,13 +164,20 @@ QucsSettingsDialog::QucsSettingsDialog(QucsApp *parent)
         "Check the schematic for common issues before running the simulation."));
     appSettingsGrid->addWidget(enableSchematicValidation, 9, 1);
 
+    appSettingsGrid->addWidget(new QLabel(tr("Validate schematic after saving:"), appSettingsTab), 10, 0);
+    validateOnSave = new QCheckBox(appSettingsTab);
+    validateOnSave->setToolTip(tr(
+        "Check the schematic for common issues after saving."));
+    appSettingsGrid->addWidget(validateOnSave, 10, 1);
+
+
     t->addTab(appSettingsTab, tr("Settings"));
 
-    appSettingsGrid->addWidget(new QLabel(tr("Set custom shortcut:"), appSettingsTab), 10, 0);
+    appSettingsGrid->addWidget(new QLabel(tr("Set custom shortcut:"), appSettingsTab), 11, 0);
     ShortcutButton = new QPushButton(appSettingsTab);
     connect(ShortcutButton, SIGNAL(clicked()),
         parent, SLOT(slotShortcutDialog()));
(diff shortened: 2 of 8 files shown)


tskaar:
Hi!

I tested this patch-set and it's a very nice QoL improvement! :)

> I added a test to catch dangling wires and unconnected ports.

IMO, this works fine as is, however down the line, I might want to hook up this kind of check to when one saves the subcircuit - Earlier detection -> Easier fix (IMO).
E.g. something like "validate schematic(s) on save", might be an option.

One nit:

image

Is it possible to add any information as a title to the issue, e.g:
`Issue #7 - Dangling Port [Warning]`, `Issue #9 - Dangling device net [Warning]`.

Makes it a bit easier to spot what kind of issue(s) one has in the schematic. 

Edit:

Also, maybe rename the button text to "OK"?
"Fix issues" might signify that there will be some automatic fixes?

andresmmera:
That's fair. Honestly, I agree with all the suggestions :-)

I'll take a look and push some commits soon.

Thank you!

andresmmera:
1) Added short title in the issue list
image

2) Added optional check after saving the schematic
image

3) Changed the text "Fix issues" by "Ok" as it suggested some automatic fixing
image

ra3xdh:
I have found a minor issue with this PR. If schematic has an unconnected ground, the validator shows a warning "Port of is not connected". The component name is empty and the warning may be misleading. I would suggest to replace it with "Ground is not connected" if the device is ground. Here is an example. 

image

andresmmera:
Thanks for pointing this out!

The problem was that GND components doesn't have a reference. 

This is now fixed by checking if the component->name is empty and, if so, use the model name instead.

image

[SchematicValidation_prj.tar.gz](https://github.com/user-attachments/files/30907475/SchematicValidation_prj.tar.gz)

ra3xdh:
Yes, the issue with GND is resolved now. Merging.

PR #1701, opened 2026-08-06
Add style properties to wires

andresmmera:
**Motivation**

I think it would be convenient to have ways to highlight nodes in the circuit. It would help to identify identify the node domains (DC, RF, etc.)

image

**Description**

This PR adds style properties to wires (color and linewidth). These properties are propagated across the whole electrical net, including nets split across multiple wire sections sharing a common name.

image

The style settings were put in the same dialog as the label name (LabelDialog, renamed as NetPropertiesDialog for more clarity). The initial node voltage is now visible: It was hidden before this PR (the user needed to click the "More" button. Honestly, I didn't know about this feature and I think it's useful that it's visible).

image

This dialog is accessible by double-clicking the net or through the net properties entry in the context menu.

image

**Persistance**
 Style is persisted to the schematic file, but only when it differs from the default style (darkBlue, linewidth=2). This makes this PR backward-compatible

image

I've used this project for testing this feature: 

[WireStyling_prj.tar.gz](https://github.com/user-attachments/files/30771877/WireStyling_prj.tar.gz)


ra3xdh:
This would be a nice feature for the next release. Merging.

andresmmera:
Thanks!

PR #1705, opened 2026-08-11
Fix issue #1703: "Reset to default" feature not working in Qucs-S Settings dialog

### LINKED ISSUE #1703: Settings Dialog: The "reset to default values" action is not working with boolean settings
The "Reset to default" button in the Qucs-S Settings dialog doesn't work well with some boolean settings

**Steps to reproduce**

1) Open the Settings dialog (CTRL + ,)
Image

2) Modify the checkboxes

Image

3) Hit "Apply" and then "OK"
4) Open the Settings dialog again (CTRL + ,)

Image

5) Hit the "Reset to Default" button -> Some checks are reset and others aren't

Image

This happens because in QucsSettingsDialog::slotDefaultValues() some checkboxes are reset manually, and others read its default value from m_Default (see [1]), and the function used for reading this fails to get the default value:

```
  template<class T>
  T itemDefault(const QString& key)
  {
    return value(key).value<T>(); // This should be return m_Defaults[key].value<T>(); Otherwise, it reads the current value!
  }
```

[1] https://github.com/ra3xdh/qucs_s/blob/current/qucs/dialogs/qucssettingsdialog.cpp#L989C1-L998C44

andresmmera:
This PR fixes issue #1703

1) It was found that the settingsManager::itemDefault [1] did not retrieve the default value, but the current value. This prevented some settings from being reset to their default value.

2) It was also found that the reset action for some settings was hardcoded, whereas others actually looked at its value in the m_Default map. The lines were also rearranged to match the same order in which the settings are presented in the dialog. The tooltips of the checkboxes were copied as comments in QucsSettingsDialog::slotDefaultValues() and settingsManager::initDefaults() for convenience. This helps to see the correspondance between values.

[1] https://github.com/ra3xdh/qucs_s/blob/current/qucs/settings.h#L23

### DIFF
--- a/qucs/dialogs/qucssettingsdialog.cpp
+++ b/qucs/dialogs/qucssettingsdialog.cpp
@@ -988,14 +988,45 @@ void QucsSettingsDialog::slotDefaultValues()
 
     undoNumEdit->setText("20");
     editorEdit->setText(QucsSettings.BinDir + "qucs");
-    checkWiring->setChecked(false);
-    allowFlexibleWires->setChecked(_settings::Get().itemDefault<bool>("AllowFlexibleWires"));
-    enableSchematicValidation->setChecked(_settings::Get().itemDefault<bool>("EnableSchematicValidation"));
-    validateOnSave->setChecked(_settings::Get().itemDefault<bool>("ValidateOnSave"));
-    checkLoadFromFutureVersions->setChecked(false);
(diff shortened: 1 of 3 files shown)


ra3xdh:
A conflict has appeared after merging #1704. Please fix before merge.

andresmmera:
Rebased. Thank you, @ra3xdh

PR #1706, opened 2026-08-15
Color gradient trace

andresmmera:
When working with Smith Charts and polar plots I find useful to display a trace using a color gradient, so that one can quickly identify high/low frequencies without placing markers.

This PR adds an option to color a trace depending on the independent variable, instead of using a plain color. Although it was intially targetted to polar and Smith Charts, this mode is available to all trace diagrams.

image

image

An extra field is added when saving a trace to a file. Backward compatibility is guaranteed
image

I tested this feature using this project:
[ColorGradientTraces_prj.tar.gz](https://github.com/user-attachments/files/31097282/ColorGradientTraces_prj.tar.gz)


ra3xdh:
Thanks! Merged.

PR #1707, opened 2026-08-15
Use images in subcircuit symbols

andresmmera:
PR #1403 introduced the ImageComponent, which allows images to be displayed in the schematic. This PR extends this functionality to support using images in subcircuit symbols.

For doing this, it was added an image-painting primitive to the qucs namespace. This also simplifies a little bit the implementation of the ImageComponent itself.

In addition to this, a SVG symbol gallery is provided so that the user can easily pick a predefined SVG symbol when creating a subcircuit. These symbols are placed at library/symbols/SVG (which after the installation are copied into usr/local/share/qucs-s/symbols/SVG). The symbols are arranged by category into folders. The user can also create its own SVG symbol library (which by default is in ~/QucsWorkspace/symbols)

image

image

The SVG gallery can be accessed from the ImageComponent properties:
image

A number of RF standard blocks were added:
image

Here's the project I've used for testing this feature:
[Image_Symbols_prj.tar.gz](https://github.com/user-attachments/files/31095090/Image_Symbols_prj.tar.gz)


ra3xdh:
Yes, this would be a useful addition for the next release. Merging this PR. The next step may be to attach SVG symbol to "SPICE library device" component. An open question is how to assign pins in such case.

PR #1715, opened 2026-08-29
Graph white background

andresmmera:
**Motivation**
Sometimes it looks better to have a white solid background in the diagrams.

**Changes**
1) An optional setting is added to the Diagram class to set the background color of the diagram to white. It was file persistance.

2) Once there, it was added an option in the context menu to copy the diagram as a png image into the system clipboard. This is useful to copy the graphs to another document without the need of taking an screenshot.
image

A checkbox is added to the diagram properties. The white background is disabled by default.
image

Here's the schematic I've used for testing this feature
[test_whiteBG.sch.tar.gz](https://github.com/user-attachments/files/31102948/test_whiteBG.sch.tar.gz)

### DIFF
--- a/qucs/diagrams/diagram.cpp
+++ b/qucs/diagrams/diagram.cpp
@@ -115,13 +115,17 @@ void Diagram::paintDiagram(QPainter *painter) {
     painter->translate(cx, cy);
 
     if (whiteBackground) {
+      int bx1, by1, bx2, by2;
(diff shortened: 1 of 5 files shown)


ra3xdh:
Yes, the "copy to image" function was missing in Qucs. It will be especially useful when writing reports using word processors. Merging this patch.

PR #1716, opened 2026-09-01
schematicValidation: reduce on-save checks to structural validation only

tskaar:
## What

This removes checks for when saving symbols (currently there are no explicit symbol checks anyways).

Additionally, on-save validation now runs only structural checks, skipping simulation-specific checks. 
Pre-simulation validation continues to run all checks as before.

## Why

The amount of "nagging" got distracting when modifying symbols, but also when modifying subcircuits (circuits that are not to be simulated standalone), where the "no simulation blocks found" error would pop-up every time one saved.

## Notes

Ideally, the simulation block check should be skipped on subcircuits, however I don't think there is an easy qualifier (at this moment anyway) to see the difference between a subcircuit block and a general schematic that is missing a simulation block

A couple of qualifiers I was considering was:

1. If there is no sources in the schematic, assume it's a subcircuit
2. If there are ports (Pins) in the schematics, assume it's a subcircuit

However, both of these have edge-cases, which then would end up being a false-negative trigger for schematic validation... So I thought just reducing the scope of `ValidationOnSave` was the safest in-between solution here.

CC: @andresmmera, since you are the implementer of the SchematicValidator, do you mind taking a look (especially on 5ea551591772f4bb901d9f458ea5bf422b5caf03)?

Thanks!

### DIFF
--- a/qucs/qucs.cpp
+++ b/qucs/qucs.cpp
@@ -3807,6 +3807,11 @@ bool QucsApp::runSchematicChecks(QWidget *w, bool isPreSimulation)
     return false;
   }
 
+  if (sch->getSymbolMode()) {
+    // There is no symbol checks; skip
+    return false;
+  }
+
   // Get the backend simulator
   QString backend = simulatorsCombobox->currentText().toLower();
 
--- a/qucs/qucs.cpp
+++ b/qucs/qucs.cpp
@@ -3826,7 +3826,14 @@ bool QucsApp::runSchematicChecks(QWidget *w, bool isPreSimulation)
 
   a_validator.setSimulationBackend(backend);
   a_validator.setSchematic(sch);
-  QVector<ValidationIssue> issues = a_validator.validate();
+  QVector<ValidationIssue> issues;
+  if (isPreSimulation) {
+    // run every check
+    issues = a_validator.validate();
+  } else {
+    // run schematic checks only
+    issues = a_validator.saveValidate();
+  }
   if (issues.isEmpty())
     return false;
 
--- a/qucs/schematicvalidator.cpp
+++ b/qucs/schematicvalidator.cpp
@@ -15,6 +15,16 @@ QVector<ValidationIssue> SchematicValidator::validate() const {
   return runAllChecks(visitedFiles);
 }
 
+QVector<ValidationIssue> SchematicValidator::saveValidate() const {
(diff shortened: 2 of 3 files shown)


andresmmera:
Sure! I'll try to take a look at this later today!

andresmmera:
@tskaar I've reviewed this PR and I tested it locally.

The functionality is good, the "soft check" when saving removes clutter.

The patch is very clean and it works as expected:
- [x] No checks when editing symbols.
- [x] No simulation-related checks when saving.
- [x] It works with the test project included in PR #1700
- [x] There are no new warnings when building.

IMHO, this PR is fine.

Regarding how to identify subcircuits, I think that the best approach is "2." since subcircuits must have pins. I can't think of any other unique feature of the subcircuits.

The alternative you mention of looking for sources, "1.", may fail because Qucs-S allows to set the initial voltage of a node (without a source). Then, it's possible that a top-level simulation won't any sources. 

See the attachment (and yes, this is a corner case...)

image

[test1.sch.tar.gz](https://github.com/user-attachments/files/31705087/test1.sch.tar.gz)

tskaar:
@andresmmera

Great, thanks for the review!

>  I think that the best approach is "2." since subcircuits must have pins. I can't think of any other unique feature of the subcircuits.

Yeah, I was thinking there _might_ someone that put pins in a subcircuit, but also wants to simulate it at that level, which _could_ happen (even though, I myself doesn't do it), so it gets a bit hairy wrt. edge-cases

andresmmera:
Well, I do exactly that in large projects for documentation purposes: If you put a pin in a top-level schematic, you can assign it a symbol. This allows you to place top-level schematics in a main page (without simulation blocks) and navigate through the project. It looks less chaotic, but it is quite a niche feature.

See, for example, this project:

https://codeberg.org/andresmmera/Open-RF-Circuit-Design/src/branch/main/RF/Filters/RF_Filter_Tutorial_prj

image

tskaar:
Ahh, that's quite a neat trick! :)

ra3xdh:
Is this PR ready for merge?

tskaar:
> Is this PR ready for merge?

Good to go from my side!

There might be some more cases I want to add for schematic testing, but that would be in a separate PR

ra3xdh:
Merged.

PR #1721, opened 2026-09-05
Monte Carlo simulation: Part 1

### LINKED ISSUE #1476: Add MonteCarlo simulation
Monte Carlo simulation is only possible using a Nutmeg script. I propose to add a special simulation type that will automatize Monte Carlo script generation. An example of Monte Carlo implementation with Nutmeg script is attached. 

[LC_band_pass_filter.sch.gz](https://github.com/user-attachments/files/22514416/LC_band_pass_filter.sch.gz)

Image

viticleri:
## Summary

Add a dedicated Monte Carlo simulation component for ngspice.

The new `.MC` component:
- targets an existing AC or SP simulation;
- generates the required Nutmeg `dowhile` loop automatically;
- issues `reset` before each run so native ngspice random expressions such as
  `unif()` and `gauss()` are re-evaluated;
- accumulates runs using the existing multi-plot result handling;
- provides `Sim` and `Runs` properties;
- prevents multiple active Monte Carlo simulations from targeting the same
  simulation;
- prevents Monte Carlo and Parameter Sweep from targeting the same simulation.

The simulation selector is limited to AC and SP analyses and excludes targets
already used by another active Monte Carlo or Parameter Sweep.

## Validation

- AC Monte Carlo: 10 runs accumulated as `Number=10`, with randomized component
  values re-evaluated between runs.
- SP Monte Carlo: 10 runs accumulated as `Number=10`, including S/Y/Z results.
- `Runs <= 0` falls back to one run.
- Duplicate `.MC` targets are rejected.
- `.MC` + `.SW` targeting the same simulation is rejected.
- Unconfigured `.MC` / `.SW` components with an empty `Sim` do not produce
  false conflicts.
- Ordinary SP simulation without `.MC` continues to run normally.

OSDI/Verilog-A model-loading issues discussed in #1476 are outside the scope of
this change.

Closes #1476


viticleri:
I noticed one small UI polish issue after opening the PR: the new Monte Carlo component currently reuses the Parameter Sweep icon, so it is displayed as "Swp" in the simulations palette.

I plan to add a dedicated "MC" icon in a small follow-up commit to this PR. No functional behavior will change.

viticleri:
Follow-up completed in c95065bb: the Monte Carlo component now has its own "MC" icon instead of reusing the Parameter Sweep icon.

This is a UI-only change; no functional behavior was modified. All CI checks are passing.

ra3xdh:
Thanks for the interest in implementing new features for Qucs-S. But this PR is not ready for merge yet. I have marked it as WIP. In the current form this PR is only a draft for MC simulation, but doesn't implement the simulation itself. I guess the AI assistance was used for preparing this PR, or you may forgot to commit the local work by mistake, because the PR is only a stub for a new feature..

Here are my specific comments. The PR could be prepared for merge if everything pointed below will be corrected. 

* The MC simulation block defines only child simulation and runs number. Where the variable at which the variance is applied could be defined? It should be possible to apply MC for component values and variables in .PARAM blocks. Also the variable scattering should be defined too. Please look at the example script in #1476 how the variance should be applied. The presented approach only wraps the simulation in the while loop:
```
let mc_mc1 = 0
dowhile mc_mc1 < 10
reset
SP DEC 100 950K 15MEG
let dBS21 = dB(S_2_1)
let dBS11 = dB(S_1_1)
write spice4qucs.sp1.plot S_1_1 Y_1_1 Z_1_1 S_1_2 Y_1_2 Z_1_2 S_2_1 Y_2_1 Z_2_1 S_2_2 Y_2_2 Z_2_2 dBS21 dBS11
set appendwrite
let mc_mc1 = mc_mc1 + 1
end
unset appendwrite
destroy all
reset
```
* The variation type `gauss()` and `unif()` are mentioned in `MonteCarlo` class description but never applied.
* MC may be applied not only to AC and SP, but for any simulation except Par.Sweep. Here is example for LTspice: https://spiceman.net/ltspice-monte-carlo-analysis/ Ngspice may use the similar approach. 
* It is a good practice to add an example for the simulation. Put it in the `examples` subdirectory in the source tree.
* Also it is a good practice to put the screenshots showing how the new feature works when submitting PR.

viticleri:
Thanks for the detailed review. You are right: the current PR only implements
the loop/result-accumulation skeleton, but it does not yet define or apply the
Monte Carlo variation itself. I had treated manually entered ngspice random
expressions as sufficient, which does not satisfy the intended feature.

I went back to the original example from #1476 and reproduced the relevant
ngspice behavior independently with ngspice 44.2.

For component values, the original approach works as expected. Nutmeg helpers
have to be defined using sunif()/sgauss(), for example:

    define unif(nom,rvar) (nom + (nom*rvar) * sunif(0))
    define gauss(nom,rvar,sig) (nom + (nom*rvar)/sig * sgauss(0))

and then applied inside the MC loop with:

    alter c1 = unif(...)
    alter l1 = gauss(...)

For .PARAM variables, ngspice accepts the corresponding functions directly:

    alterparam p = unif(...)
    alterparam p = gauss(...)
    reset

I also tested aunif(), agauss() and limit() in both paths. For component
values the Nutmeg helper definitions are required, while alterparam can use
the ngspice parameter functions directly.

One important detail is the ordering when both .PARAM variables and component
values are varied in the same run. A reset after a device alter restores that
device to its netlist value. The working order is therefore:

    [all alterparam statements]
    reset
    [all device alter statements]
    [analysis]

Multiple alterparam statements can share a single reset. For component-only
variation no reset is needed inside the loop.

The #1476 example also shows that several quantities may vary simultaneously
in one Monte Carlo analysis. Using one .MC block per quantity would generate
nested loops, so one .MC block needs to contain a list of varied quantities.
Each entry would identify the quantity together with its distribution and
scattering parameters.

For component versus .PARAM detection, the existing Param_Sweep code provides
a useful precedent: a name resolving to a schematic component can use alter,
otherwise it can use alterparam. I would reuse that mechanism, including the
existing conversion from a Qucs component name to the ngspice reference,
unless you prefer the type to be explicit in each MC entry.

I also plan to lift the current AC/SP-only restriction. I have inspected the
different ngspice result paths, but I do not want to assume that simply
wrapping every analysis is sufficient. AC and S-parameter already have the
multi-run result path used by the current implementation. Transient uses the
same generic write path and is the natural next case to validate. Other
analyses require additional consideration: NOISE and DC have special
per-iteration output handling, DISTO relies on variables supplied by
Eqn/NutmegEq, and PZ/SENS/SENS_AC/FFT use different result paths that may
require additional collection/parser work.

Before changing the implementation, I would therefore like to confirm three
design points:

1. When you say that MC may be applied to any simulation except Parameter
   Sweep, do you expect all supported ngspice analyses to work in this PR
   before it is mergeable, including the analyses with special result paths,
   or is it acceptable to implement them incrementally?

2. For each varied quantity, should the nominal value be explicitly stored in
   the MC definition, as in the original #1476 script, or should Qucs-S obtain
   it automatically from the selected component/.PARAM where possible?

3. For multiple varied quantities, is a compact list stored in the .MC
   component acceptable for the first implementation, or do you prefer a
   structured GUI editor for the quantities/distributions?

I also noticed that Qucs-S already ships
examples/templates_ngspice/Monte_Carlo.sch using the older
SpiceFunc/SpicePar/Parameter-Sweep approach. I will add an example for the new
.MC component as requested; please let me know whether you would prefer the
existing template to remain as an alternative example or eventually be
replaced.

I will keep the implementation minimal and reuse the existing
Param_Sweep/ngspice mechanisms rather than introducing a separate simulation
framework.

ra3xdh:
See https://github.com/viticleri/qucs_s_viticleri/pull/2 I have implemented a full MC simulation based on the provided skeleton. After merging this, the new simulation will be ready. 

image

ra3xdh (approved):

ra3xdh:
Merging this. The remaining stuff for implementing MC simulation implemented via #1727

PR #1722, opened 2026-09-07
Add substrate material selector to Transcalc

### LINKED ISSUE #354: Add integration between Substrate.lib and Transcalc tool
Transcalc tool requires substrate properties to perform the line calculation. The current implementation requires hand input of these properties. It's need to add a possibility to select the substrate material from the drop-down list. The material properties should be taken from `Substrate.lib` library.

viticleri:
## Summary

Add integration between the Qucs substrate library and Transcalc.

Transcalc now provides a material selector in the Substrate Parameters section. Selecting a material loads the compatible properties from `Substrates.lib` into the existing fields while keeping manual input fully editable.

The selector is additive:
- `Manual` remains the default option and changes nothing.
- Selecting a material performs a one-shot copy of its properties.
- Changing transmission-line type or loading a `.trc` file resets the selector to `Manual`.
- If `Substrates.lib` is unavailable, Transcalc continues to work normally with manual input.

A small standalone reader is used instead of coupling Transcalc to the main Qucs library parser.

## Property mapping

Depending on the current transmission-line type, compatible properties are applied as:

- `er` → `Er`
- `h` → `H` / `h`
- `t` → `T`
- `tand` → `Tand`
- `rho` → `Cond` / `Sigma` as `1 / rho`
- `D` → `Rough`

Properties not supported by the current mode are left unchanged.

## Tests

Added a frameworkless `SubstrateLibParser` CTest covering:
- valid substrate entries
- material names containing spaces and `/`
- numeric and unit parsing
- malformed entries
- unsupported units
- missing library files
- `rho == 0` parsing without division

Validation performed:

- 2/2 tests passed
- manual verification with Microstrip, Stripline and Coaxial
- verified manual editing remains possible after material selection
- verified selector reset on mode changes
- verified selector reset when loading `.trc` files, including same-mode loads
- verified graceful behavior when `Substrates.lib` is unavailable

Fixes #354


andresmmera:
I've taken a look at the PR. It works, but I have a few comments:

1. If the user selects a predefined substrate, it resets to “manual” when the user changes the transmission line implementation. For example, synthesize a microstrip line with RO4003 and then change it to a stripline. The material resets to “manual” instead of keeping the RO4003C values.

2. Minor issue: I think that it'd be clearer if in the combo you put “Custom” for material, instead of “Manual”. “Manual entry” sounds good to me as well.

3. About the test file, test_substratelib.cpp, I'm not sure if this is the right place to put a test file. I think there is no test code across the project.

4. Not related to the code itself, but the dielectric permittivity of RO4003C should be 3.55 [1]. This is the value they recommend for design, not 3.38.

[1] https://www.rogerscorp.com/-/media/project/rogerscorp/documents/advanced-electronics-solutions/english/data-sheets/ro4000-laminates-ro4003c-and-ro4350b---data-sheet.pdf

viticleri:
Thanks for the review. I pushed a follow-up commit addressing the first two points.

1. The selected substrate is now kept when changing the transmission-line type, and its compatible values are reapplied to the new implementation. Loading a `.trc` file still switches the material selector back to `Custom`, since the values then come from the file.

2. I changed `Manual` to `Custom`.

3. Regarding `test_substratelib.cpp`, I checked the current tree and there is an existing CTest precedent in `qucs/geometry/test_geometry.cpp`, located alongside the implementation and registered from the local `CMakeLists.txt` with `add_executable()` / `add_test()`. I followed that pattern here. If you prefer the Transcalc test to be placed elsewhere, I can move it.

4. I left the RO4003 value unchanged in this PR because `3.38` is pre-existing data in `Substrates.lib`; this PR only consumes the existing library. I think updating the substrate data would be better handled separately from the Transcalc integration.

I also re-tested the mode change and `.trc` loading behavior locally, and the new CI run has passed.

ra3xdh:
Thanks for the contribution!

I will test it in the next few days and give feedback. 

> I checked the current tree and there is an existing CTest precedent in qucs/geometry/test_geometry.cpp

This file is a leftover from the draft version and not in use. I overlooked it when accepting the patch implementing the stuff in geometry subdirectory. In general the project has no test policy and the patches has no obligation to be covered by test. The tests should be introduced by a separate patch if required. But my opinion that the obligation to write test will rather distract new contributors.

viticleri:
Thanks for the clarification.

I removed `test_substratelib.cpp` and its CTest target from this PR in commit `8172b587`, keeping the substrate integration itself unchanged.

The new CI run has passed successfully.

ra3xdh:
The new feature works as expected. Merging this. Thanks for the contribution!

PR #1727, opened 2026-09-13
Monte Carlo simulation: part 2

### LINKED ISSUE #1476: Add MonteCarlo simulation

ra3xdh:
This PR contains a full implementation of MonteCarlo simulation

* Added properties to define variable, function, and function parameters. I decided to restrict only to variables using `alterpaparam`, because using `gauss` and `unif` functions with `alter` directive always failing. There may be bug in Ngspice.
* Runs check is moved to Schematic validator
* Implemented possibility to apply MC to semicolon separated lists of variables. 
* MC is allowed for SP,AC,TRAN,DC sweep

Test schematic is attached.

[LPF40.sch.gz](https://github.com/user-attachments/files/32064026/LPF40.sch.gz)

image


ra3xdh:
This PR also contains a fix of SchematicValidator. It uses `Component::isSimulation` to determine if schematic contain a simulation.

viticleri:
I completed a post-merge validation on `current` at `99848aad`.

The intended Monte Carlo cases are working correctly:

- MC -> SP: 10 runs
- MC -> AC: 10 runs, final dataset contains `Number 10`
- MC -> TRAN: 10 runs
- MC -> DC Sweep: 10 sweeps, final dataset contains `Number 10`
- MC/Parameter-Sweep conflict is correctly detected by `SchematicValidator`

I found one remaining issue with the MC simulation selector.

The `Sim` combo currently exposes a direct `.DC` simulation (`DC1`) as a valid Monte Carlo target. I reproduced this with `MC1 -> DC1`, `Runs=10`.

Ngspice executes all 10 iterations, but the generated loop contains:

```text
op
print v(B) v(Uin) i(VPr1) i(VPr3) i(VPr4) v(Vout) > spice4qucs.dc1.ngspice.dc.print
set appendwrite
```

The `print ... >` command rewrites the same file on every iteration. After the 10 runs, `spice4qucs.dc1.ngspice.dc.print` contains only one result set, and the resulting `.dat.ngspice` file has no `<indep Number 10>` dimension.

So direct `.DC` is currently selectable through the normal GUI, but its Monte Carlo results are not accumulated.

Since the PR describes MC support as SP, AC, TRAN and DC Sweep, the minimal fix would appear to be to exclude direct `.DC` from the MC `Sim` selector. If direct `.DC` is intended to be supported as well, its output accumulation path needs separate handling from the raw-file `write` path used by the supported analyses.

